PR #70, opened 2017-10-31
Add Ellipsoid and ellipsoid sampling

rimadoma:


imagejan:
Any plans to port these to core `imagej-ops`?

rimadoma:
Yes! That's something I'm going to do at the upcoming Dresden hackathon. But first we need to determine which linalg library ImageJ should commit to. Afaik Vector3d etc from imglib2 is in some kind of limbo. This family of ops may even find it's way to a brand new 'imagej-linalg-ops' module, partly to avoid adding dependencies to 'imagej-ops'.

PR #71, opened 2017-11-01
Refactor rotation op

rimadoma:
The RotateAboutAxis op had too many bells and whistles and was too clunky.


rimadoma:
Merging a failing build, because the failure was caused by an unrelated phantom bug that very occasionally rears its ugly head.

PR #76, opened 2017-11-20
Box intersect

rimadoma:
Adds the box intersect op for finding the intersection points between a line and an interval.

### DIFF
--- a/Modern/ops/src/main/java/org/bonej/ops/BoxIntersect.java
+++ b/Modern/ops/src/main/java/org/bonej/ops/BoxIntersect.java
@@ -1,0 +1,174 @@
+
+package org.bonej.ops;
+
+import java.util.Optional;
+import java.util.stream.DoubleStream;
+
+import net.imagej.ops.Contingent;
+import net.imagej.ops.Op;
+import net.imagej.ops.special.function.AbstractBinaryFunctionOp;
+import net.imglib2.Interval;
+import net.imglib2.type.numeric.real.DoubleType;
+import net.imglib2.util.ValuePair;
+
+import org.scijava.plugin.Parameter;
+import org.scijava.plugin.Plugin;
+import org.scijava.vecmath.Tuple3d;
+import org.scijava.vecmath.Vector3d;
+
+/**
+ * Solves scalar values (t<sub>1</sub>, t<sub>2</sub>) for the parametric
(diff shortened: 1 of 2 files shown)


Modern/ops/src/test/java/org/bonej/ops/BoxIntersectTest.java
```diff
@@ -0,0 +1,337 @@
...
+			direction);
+
+		final Optional<ValuePair<DoubleType, DoubleType>> result = boxIntersect
+			.calculate(line);
+
+		assertTrue(result.isPresent());
+		final ValuePair<DoubleType, DoubleType> scalars = result.get();
+		assertEquals(Math.sqrt(2) / 2.0, scalars.a.get(), 1e-11);
+		assertEquals(0.848528137423857, scalars.b.get(), 1e-11);
+	}
```
alessandrofelder: exact value is Math.sqrt(2)*0.6

Modern/ops/src/main/java/org/bonej/ops/BoxIntersect.java
```diff
@@ -0,0 +1,174 @@
...
+		final ValuePair<Tuple3d, Vector3d> line, final Interval interval)
+	{
+		final int d = interval.numDimensions();
+		final long[] minBounds = new long[d];
+		interval.min(minBounds);
+		final long[] maxBounds = new long[d];
+		interval.max(maxBounds);
+		final Vector3d direction = new Vector3d(line.b);
+		final double length = direction.length();
+		if (length > 0.0 && Double.isFinite(length)) {
```
alessandrofelder: throw IllegalArgumentException instead of this check?
I don't think there is a meaningful way of dealing with infinite entries in the vectors.

Modern/ops/src/test/java/org/bonej/ops/BoxIntersectTest.java
rimadoma: Cheers, fixed in e4a2cef1ea31376035d905d6793845b64b8eab4c

Modern/ops/src/main/java/org/bonej/ops/BoxIntersect.java
rimadoma: Fixed in f379f79a5dc3ba904869c53e259aad86ac1b0107

PR #90, opened 2018-03-12
Fix Anisotropy calculations

rimadoma:
Fix anisotropy so that the method follows science better. MIL vector is now just one vector that's the summation of total length of lines / their interceptions. This gives more stable and correct results. 

Summary:
- Add MILPlane op and its helper class
- Remove MILGrid op and its helper classes
- Fix the wrapper
- Perform tests on the wrapper

There are commits that go back and forth adding stuff to classes that later get removed, but IMO it's worth keeping them for now, in case we need to revert design.


Modern/ops/src/main/java/org/bonej/ops/mil/MILPlane.java
```diff
@@ -0,0 +1,294 @@
...
+ * number of times <em><b>v</b></em> intercept foreground, and <em>h</em> is the
+ * sum of lengths of <em><b>v</b></em>. The lines <em><b>v</b></em> are traced
+ * through the interval. A line intercepts foreground, when the previous point
+ * sampled along it is background, and the current one is foreground.
+ * </p>
+ * <p>
+ * For example MIL vectors can be used to estimate the anisotropy of the
+ * "texture" in an image. It's best suited for images that are completely
+ * filled, and a part of a larger whole, e.g. a section of trabecular bone.
+ * </p>
```
alessandrofelder: detail: I may be wrong, but "section" implies 2-dimensionality to the physiologist in me... "volume" may be better?

Modern/ops/src/main/java/org/bonej/ops/mil/MILPlane.java
```diff
@@ -0,0 +1,294 @@
...
+	 */
+	@Parameter(required = false, persist = false)
+	private Long seed;
+
+	/**
+	 * Calculates the MIL vector of the interval.
+	 *
+	 * @param interval a 3D interval.
+	 * @param rotation direction of the MIL lines in the interval.
+	 * @return total length / total phase changes of vectors sampled in interval.
```
alessandrofelder: detail: This could be slightly more precise: possibly "a vector with magnitude total length sampled / total phase changes from background to foreground and direction of the MIL lines"

Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/AnisotropyWrapper.java
```diff
@@ -74,49 +75,51 @@
...
-	private static int ROTATIONS = 1_000;
-	private static int DEFAULT_LINES = 50;
+	/**
+	 * Default directions is 2_000 since that's roughly the number of points in
+	 * Poisson distributed sampling that'd give points about 5 degrees apart).
+	 */
+	private static int DEFAULT_DIRECTIONS = 2_000;
+	// Magic number, there are no good guesses what might be the right number of
+	// lines. Selected by fiddling around with the tool
+	private static int DEFAULT_LINES = 100;
```
alessandrofelder: maybe formulate this more positively: "Number found to be sensible from experience with data at hand, may differ for other data" or similar?

Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/AnisotropyWrapper.java
```diff
@@ -222,11 +211,9 @@ private boolean isCalibrationIsotropic() {
...
 	}
 
-	@SuppressWarnings("unchecked")
 	private void matchOps(final Subspace<BitType> subspace) {
-		milOp = (BinaryFunctionOp) Functions.binary(opService, MILGrid.class,
-			List.class, subspace.interval, new AxisAngle4d(), lines,
-			samplingIncrement, new Random());
+		milOp = Functions.binary(opService, MILPlane.class, Vector3d.class,
+			subspace.interval, new AxisAngle4d(), lines, samplingIncrement);
 		final List<Vector3d> tmpPoints = generate(Vector3d::new).limit(
```
alessandrofelder: comment: this spot may be interesting to cite and discuss in the BoneJ paper: the Moreno et al paper we refer to proposes a different way of doing MIL - how would the milOp signature for that way look?

Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/AnisotropyWrapper.java
```diff
@@ -237,35 +224,79 @@ private void matchOps(final Subspace<BitType> subspace) {
...
-		final ExecutorService executor = Executors.newFixedThreadPool(5);
-		final Callable<List<Vector3d>> milTask = () -> milOp.calculate(interval);
-		final List<Future<List<Vector3d>>> futures = Stream.generate(() -> milTask)
-			.limit(rotations).map(executor::submit).collect(toList());
+		final int cores = Runtime.getRuntime().availableProcessors();
+		// The parallellization of the the MILPlane algorithm is a memory bound
+		// problem, which is why speed gains start to drop after 5 cores. With much
+		// larger 'nThreads' it slows down due overhead. Of course '5' here is a bit
+		// of a magic number, which might not hold true for all environment, but we
+		// need some kind of upper bound
```
alessandrofelder: typos: "due **to** overhead" and "for all environment**s**

Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/AnisotropyWrapper.java
rimadoma: Let's discuss this. I'm not entirely sure what you mean.

PR #91, opened 2018-03-13
Remove the Help command

rimadoma:
Thanks to the new search bar in ImageJ, the plug-in that opens user documentation web pages has become redundant. The info will be on the ImageJ wiki and forum, where the search bar can find it.


rimadoma:
I added a mention to the PR description above about the formatting.

PR #92, opened 2018-03-13
Delete testImgs module

rimadoma:
Deletes the largely redundant testImgs module. Modifies tests so that they're self contained, and don't require an additional module that just hangs around. Manual tests should be repeatable on an installation from the BoneJ update site, which excludes testImgs, because that module is not hosted there.


Modern/wrapperPlugins/src/test/java/org/bonej/wrapperPlugins/ElementFractionWrapperTest.java
```diff
@@ -65,43 +61,15 @@ public void testNonBinaryImageCancelsElementFraction() throws Exception {
...
-
-		assertTrue("A non 2D and 3D image should have cancelled the plugin", module
-			.isCanceled());
-		assertEquals("Cancel reason is incorrect", CommonMessages.WEIRD_SPATIAL,
-			module.getCancelReason());
-		verify(mockUI, timeout(1000)).dialogPrompt(anyString(), anyString(), any(),
-			any());
-	}
-
+	// TODO Rewrite with a hyper stack
```
alessandrofelder: As discussed, a hyper-stack should probably report BV/TV for each channel at each time point.

rimadoma: Cheers for the review

PR #94, opened 2018-05-24
Legacy clean

rimadoma:
Addresses different style issues and "code smells" in effort to make the code easier to understand and maintain. Some style issues are addressed just to make code more consistent across modules.  Also fixes licences to BSD-2. `pom-scijava` is updated to 20.1.0 but no further due a build [issue](https://github.com/imagej/imagej-maven-plugin/issues/30).


rimadoma:
It's annoying that the update in https://github.com/bonej-org/BoneJ2/pull/94/commits/cd0051ea9ae750dbc4a989771c0e767c738d4892 breaks `mvn install` for now, but it's more important to depend on an up-to-date version of `pom-scijava`.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleCounter.java
```diff
@@ -2204,23 +2136,25 @@ private void joinStructures(final ImagePlus imp, final int[][] particleLabels,
...
 		final int z0 = voxel[2];
-		if (phase == BACK) {
-			return Stream.of(new int[] { x0 - 1, y0, z0 }, new int[] { x0 + 1, y0,
-				z0 }, new int[] { x0, y0 - 1, z0 }, new int[] { x0, y0 + 1, z0 },
-				new int[] { x0, y0, z0 - 1 }, new int[] { x0, y0, z0 + 1 }).filter(
-					a -> withinBounds(a[0], a[1], a[2], w, h, d)).collect(Collectors
-						.toList());
-		}
+		return get26NeighbourhoodCoordinates(x0, y0, z0, bounds[0], bounds[1], bounds[2]);
+	}
```
alessandrofelder: why not `return get26NeighbourhoodCoordinates( voxels[0],voxels[1],voxels[2],bounds[0],...` ?

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleCounter.java
```diff
@@ -2230,6 +2164,25 @@ private void joinStructures(final ImagePlus imp, final int[][] particleLabels,
 		return coordinates;
 	}
 
+	private static List<int[]> get6NeighbourhoodCoordinates(final short[] voxel,
+													 final int[] bounds)
+	{
+		final int x0 = voxel[0];
+		final int y0 = voxel[1];
+		final int z0 = voxel[2];
+		return get6NeighbourhoodCoordinates(x0, y0, z0, bounds[0], bounds[1], bounds[2]);
```
alessandrofelder: why not `return get26NeighbourhoodCoordinates( voxels[0],voxels[1],voxels[2],bounds[0],...` ?

Legacy/bonej/src/main/java/org/bonej/plugins/EllipsoidFactor.java
```diff
@@ -144,15 +140,14 @@ public void run(final String arg) {
 		gd.addNumericField("Gaussian_sigma", 2, 0, 4, "px");
 		gd.addCheckbox("Flinn_plot", false);
 
-		gd.addMessage("Ellipsoid Factor is beta software.\n" + "Please report your experiences to the user group:\n"
-				+ "http://bit.ly/bonej-group");
+		gd.addMessage("Ellipsoid Factor is beta software.\n" +
+			"Please report your experiences to the user group:\n" +
+			"http://bit.ly/bonej-group");
 		gd.addHelp("http://bonej.org/ef");
```
alessandrofelder: should this not point to the imagej forum these days?

Legacy/bonej/src/main/java/org/bonej/plugins/Orienteer.java
```diff
@@ -224,38 +226,263 @@ public Orienteer() {
...
+	}
+
+	/**
+	 * Get the orientation of the principal direction
+	 *
+	 * @param imp an image.
+	 * @return Orientation in radians clockwise from 12 o'clock
+	 */
+	private double getOrientation(final ImagePlus imp) {
+		final Integer id = imp.getID();
```
alessandrofelder: 3 implementations of `getOrientation` do pretty different things depending on signature

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleCounter.java
```diff
@@ -410,48 +441,337 @@ else if (choice.equals(items2[1]))
...
+		}
+	}
+
+	private static void displayEllipsoids(final Object[][] ellipsoids,
+		final Image3DUniverse univ)
+	{
 		final int nEllipsoids = ellipsoids.length;
-		ellipsoidLoop: for (int el = 1; el < nEllipsoids; el++) {
+		for (int el = 1; el < nEllipsoids; el++) {
 			IJ.showStatus("Rendering ellipsoids...");
```
alessandrofelder: why does this loop start at 1 ?????

rimadoma: I don't know exactly, but it seems to consistent in other parts of the code where the `ellipsoid[][]` is looped. Maybe `ellipsoid[0]` is the background particle...

Legacy/bonej/src/main/java/org/bonej/plugins/EllipsoidFactor.java
rimadoma: Good catch

Legacy/bonej/src/main/java/org/bonej/plugins/Moments.java
```diff
@@ -422,39 +270,45 @@ private ImagePlus alignToPrincipalAxes(final ImagePlus imp, final Matrix E, fina
...
+		else if (hi <= wi && wi <= di) {
 			IJ.log("Case 2");
 			rotation = E.times(RotZ);
-		} else if (di <= hi && hi <= wi) {
+		}
+		else if (di <= hi && hi <= wi) {
 			IJ.log("Case 3");
 			rotation = E.times(RotY);
-		} else if (hi <= di && di <= wi) {
+		}
```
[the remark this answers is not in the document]
rimadoma: That's indeed the case.

Legacy/bonej/src/main/java/org/bonej/plugins/Moments.java
```diff
@@ -435,10 +435,10 @@ private ImagePlus alignToPrincipalAxes(final ImagePlus imp, final Matrix E, fina
 		} else if (di <= hi && hi <= wi) {
 			IJ.log("Case 3");
 			rotation = E.times(RotY);
-		} else if (hi <= di && di <= wi) {
+		} else if (hi <= di) {
 			IJ.log("Case 4");
 			rotation = E.times(RotY).times(RotZ);
-		} else if (di <= wi && wi <= hi) {
+		} else if (wi <= hi) {
 			IJ.log("Case 5");
```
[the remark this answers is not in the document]
rimadoma: Else if - blocks are dependent on each other, you only advance if all the previous statements have been false. Maybe another way of writing these statements would be clearer, e.g. 
```java
if (wi<=hi) {
  if (hi <= di) {
       ...
  }
  else {
    if (di <= wi) {
       ...
    }
    else {
       ...
    }
  } 
} else {
       ...
}
```

Legacy/bonej/src/main/java/org/bonej/plugins/Orienteer.java
```diff
@@ -71,15 +72,15 @@
 	private int axis0 = 1;
 
 	/** Current secondary direction choice */
-	private int axis1 = 0;
+	private int axis1;
```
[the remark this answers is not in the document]
rimadoma: IMO yes, because in Java variables do have default values unlike in other languages. Initializing primitives zero or references null is redundant. Also only initializing only the variables that don't have default values directs your attention better.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleCounter.java
rimadoma: Rather I think this should be `(x0, y0, z0, w, h, d)`

PR #97, opened 2018-06-07
Anisotropy result fix

rimadoma:
Fixes Anisotropy results so that either all or none are shown in the results table. More specifically, this fixes an issue where fitting would succeed for some and fail for other 3D subspaces of a hyperstack. Then when the user ran another BoneJ plug-in successfully the result table would show these partial results.

There are no new tests for the fix, because I wasn't able to devise a repeatable manual or automatic test.

NB: This branch was created from the `up-pom-scijava` branch on purpose.

### DIFF
--- a/Modern/wrapperPlugins/pom.xml
+++ b/Modern/wrapperPlugins/pom.xml
@@ -156,7 +156,7 @@
             <artifactId>imagej-mesh</artifactId>
         </dependency>
         <dependency>
-            <!-- Needed legacy things like ImagePlus UI. Do not remove. -->
+            <!-- Needed for legacy things like ImagePlus UI. Do not remove. -->
             <groupId>net.imagej</groupId>
             <artifactId>imagej-legacy</artifactId>
             <scope>runtime</scope>
--- a/Modern/wrapperPlugins/pom.xml
+++ b/Modern/wrapperPlugins/pom.xml
@@ -142,6 +142,7 @@
         <dependency>
             <groupId>net.imagej</groupId>
             <artifactId>imagej-ops</artifactId>
+            <version>${ij-ops.version}</version>
         </dependency>
         <dependency>
             <groupId>net.imagej</groupId>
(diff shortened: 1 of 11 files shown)


Modern/wrapperPlugins/pom.xml
```diff
@@ -151,6 +151,10 @@
             <groupId>net.imagej</groupId>
             <artifactId>ij1-patcher</artifactId>
         </dependency>
+        <dependency>
+            <groupId>net.imagej</groupId>
+            <artifactId>imagej-mesh</artifactId>
+        </dependency>
         <dependency>
             <!-- Needed legacy things like ImagePlus UI. Do not remove. -->
             <groupId>net.imagej</groupId>
```
alessandrofelder: typo: Needed **for** legacy things...

Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/AnisotropyWrapper.java
```diff
@@ -146,23 +146,25 @@ public void run() {
...
 		matchOps(subspaces.get(0));
-		// TODO Does it make more sense to collect all the results we can (without
-		// cancelling) and then report errors at the end?
+		final List<Ellipsoid> ellipsoids = new ArrayList<>();
 		for (int i = 0; i < subspaces.size(); i++) {
 			statusService.showStatus("Anisotropy: sampling subspace #" + (i + 1));
 			final Subspace<BitType> subspace = subspaces.get(i);
-			if (!processSubspace(subspace)) {
+			final Ellipsoid ellipsoid = foo(subspace);
+			if (ellipsoid == null) {
```
alessandrofelder: `foo` seems like a too general method name to me?

rimadoma: Oops - good catch - not indeed a good name. Must've forgotten to change that.

PR #99, opened 2018-06-08
Modern clean

rimadoma:
Simplifies code and fixes style and formatting issues in the `Modern` module


Modern/ops/src/main/java/org/bonej/ops/CleanShortEdges.java
```diff
@@ -313,8 +314,11 @@ private void euclideanDistance(final Edge e) {
 	}
 
 	private double calibratedLength(final Vector3d centre1) {
-		return Math.sqrt(Math.pow(centre1.x * calibration3d.get(0), 2) + Math.pow(centre1.y * calibration3d.get(1), 2)
```
alessandrofelder: `centre1` is a weird name for a vector whose length we want to calculate? Maybe it should be called `uncalibratedVector3d`?

Modern/utilities/src/test/java/org/bonej/utilities/SharedTableTest.java
```diff
@@ -106,7 +106,7 @@ public void testAddLabelRepeatedly() throws Exception {
 	}
 
 	@Test
-	public void testFoo() throws Exception {
+	public void testFoo() {
```
alessandrofelder: Is this a good name?

Modern/ops/src/main/java/org/bonej/ops/CleanShortEdges.java
rimadoma: Good catch, but this PR only affected the formatting. In any case, PR #103 deletes this method.

PR #103, opened 2018-06-19
Call ITA methods from AnalyzeSkeleton_

rimadoma:
Calls methods used in the ITA wrapper from _sc.fiji.AnalyzeSkeleton__ instead of ops. These methods don't generalize well into ops, as they depend on many idiosyncrasies of _AnalyzeSkeleton__. Thus it's better to implement them as simple utility methods. This PR also removes the now redundant ops.

This PR
* Removes ITA related ops and utilities
* Calls helper classes from `sc.fiji.AnalyzeSkeleton_` (added there by us)
* Updates manual tests
* Updates POMs

See [PR to AnalyzeSkeleton](https://github.com/fiji/AnalyzeSkeleton/pull/12).


rimadoma:
I squashed the commit history, because there were broken commits. Maven build failed, because `AnalyzeSkeleton_:3.3.0` requires `joml:1.9.x`, which in turn has `JDK 1.9` bytecode. Newer versions of `pom-scijava` can handle 1.9, but this branch had a dependency on too old a version when it started.

PR #113, opened 2018-08-08
Put all the BoneJ plugins in the same menu

mdoube:
Legacy and IJ2 plugins are separated by a line. Whether a plugin is
legacy or not should be relatively invisible to end users.

### DIFF
--- a/Legacy/bonej/src/main/resources/plugins.config
+++ b/Legacy/bonej/src/main/resources/plugins.config
@@ -21,19 +21,19 @@
 ###
 # Author: Michael Doube
 
-Plugins>BoneJ Legacy>Analyze, "Orientation", org.bonej.plugins.Orienteer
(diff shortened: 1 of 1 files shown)


rimadoma:
Fair enough. My reasoning was that the legacy and IJ2 plug-ins have slightly different look and feel, e.g. different results table. Also I wanted a plug-in moving from _BoneJ  Legacy_ to _BoneJ_ underline to the user: _the plug-in has changed: check your results and scripts_.

imagejan:
> e.g. different results table

That's an important point. Currently, the IJ1<>IJ2 synchronization isn't perfect with regards to results tables (see also https://github.com/imagej/imagej-legacy/issues/189). Users should expect to be able to do the exact same things with IJ2 tables as with IJ1 `ResultsTable`s, i.e. saving them to file, accessing them from scripts, ...

PR #119, opened 2018-08-13
Fix result inserter logic (Legacy)

mdoube:
Result inserter works more intuitively if a bunch of new results are added to the last line rather than the first line a duplicate entry is found.


rimadoma:
Thanks!

PR #133, opened 2018-09-04
Thickness: add all the results to the same row

### LINKED ISSUE #132: Thickness produces extra rows in the result table
**Describe the bug**
When Thickness completes, it adds two new rows to the result table:
ImageTitle_Tb.Th
Image Title_Tb.Sp

**To Reproduce**
1. Open an image
2. Run Plugins > BoneJ > Thickness

**Expected behavior**
Results should be entered in the row with label according to the original input image

**Screenshots**
screenshot from 2018-09-04 09-31-51

**Environment**
 - OS: Ubuntu 16.04
 - Java version: 1.8.0_172 (Oracle)
 - Fiji updated today
- BoneJ2 5e3be0cd69b833fa06a71be3a85532ab597b8441

mdoube:
Also update tests to reflect single row of results

Fixes #132

### DIFF
--- a/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/ThicknessWrapper.java
+++ b/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/ThicknessWrapper.java
@@ -132,7 +132,7 @@ public class ThicknessWrapper extends ContextCommand {
 			final ImagePlus map = createMap();
 			statusService.showStatus("Thickness: calculating results");
 			final String label = inputImage.getTitle();
-			addMapResults(label, map);
+			addMapResults(map);
 			thicknessMaps.put(foreground, map);
 		});
 		if (SharedTable.hasData()) {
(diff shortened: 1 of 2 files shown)


Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/ThicknessWrapper.java
```diff
@@ -143,12 +144,11 @@ public void run() {
 		}
 	}
 
-	private void addMapResults(final ImagePlus map) {
+	private void addMapResults(String label, final ImagePlus map) {
 		if (map == null) {
```
rimadoma: The parameter isn't necessary. Because `inputImage` is a member of the class, this non-static method can get the label by calling `inputImage.getTitle()`

Modern/wrapperPlugins/src/test/java/org/bonej/wrapperPlugins/ThicknessWrapperTest.java
```diff
@@ -198,8 +198,8 @@ public void testResults() throws Exception {
 		for (int i = 0; i < 6; i++) {
 			final DefaultColumn<Double> column = table.get(i);
 			assertEquals(expectedHeaders[i], column.getHeader());
-			assertEquals("Results table has wrong number of rows", 2, column.size());
-			for (int j = 0; j < 2; j++) {
+			assertEquals("Results table has wrong number of rows", 1, column.size());
+			for (int j = 0; j < 1; j++) {
```
rimadoma: Nested loop is redundant now that we have only one row (no need for iteration).

rimadoma:
Cheers!

Issue #115, opened 2018-08-09
Show progress bar when wrappers are running


rimadoma:
For each wrapper plug-in show the progress bar with `statusService.showProgress(completed, tasks);` in an appropriate manner. If the tasks are quick you could show progress with "1 / n subspaces processed". Otherwise you could show progress for each sub-space with "1 / n  processing completed" like in `Anisotropy`.

ctrueden:
I think there is opportunity to improve the SciJava core here. The `StatusService` is a singleton progress bar and message, same as ImageJ1. There is now a `TaskService` that lets you spawn new tasks and keep track of their progress individually. What I have wanted for a long time is to tie in the tasks to the `StatusService`, so that you can do something like:
```java
Task task = taskService.createTask("Processing subspaces");
for (int i=0; i<subspaceCount; i++) {
  Task subtask = taskService.createTask("Processing subspace #" + i);
  task.addSubtask(subtask);
}
subspaceProcessor.launch(task); // or launch each subtasks, if multithreaded
statusService.setTask(task);
```

And then the status bar would reflect the status of the running tasks, including messages and progress, automatically. Some day, the ImageJ UI could even be improved to provide a dropdown details view similar to Eclipse with its subtasks, so that you can see everything currently going on.

The above is very rough off the top of my head—see the [`org.scijava.download` package](https://github.com/scijava/scijava-common/tree/scijava-common-2.75.0/src/main/java/org/scijava/download) where I really used the `TaskService` in practice.

The two things needed to make the above a reality:

1. Adding API to the `StatusService` to tie in with `Task`s.
2. Add API to the `Task` to allow for nested tasks.

What do you think @rimadoma? Does it make sense?

rimadoma:
@ctrueden That sounds very handy, but frankly here I'm just trying to make the UI look like something is happening, so that the user doesn't BoneJ has frozen. In any case most of the action happens inside ops, so the bar wouldn't be updated that often even it worked in finer detail than per subspace. 

Is there a way to create a progress bar for a task with unknown number of iterations / time? Just something that darts back and forth while processing happens.

ctrueden:
> Is there a way to create a progress bar for a task with unknown number of iterations / time?

No. ImageJ1 uses `java.awt.ProgressBar` which does not have an indeterminate mode, unlike Swing's `JProgressBar`. And we are stuck with the ImageJ1 UI.

imagejan:
[...]

PR #136, opened 2018-09-10
Add progress bars

rimadoma:
* Fixes issue #115 
* `AnalyseSkeletonWrapper`, `SkeletoniseWrapper` and `ThicknessWrapper` don't have calls to the progress bar, because the underlying plug-ins use it themselves.
* `IntertrabecularAngleWrapper` doesn't call progress bar, because it's too fast for the bar to really show. Also it calls `AnalyseSkeletonWrapper` and `SkeletoniseWrapper` which use the bar. That is, user sees something happening already.
* Adds manual tests. It'd be better to mock `StatusService` and inject into the ImageJ context in the unit tests. Added a note to the [back log](https://github.com/bonej-org/BoneJ2/wiki/Backlog)


Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/IntertrabecularAngleWrapper.java
```diff
@@ -88,7 +88,7 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
  * @see sc.fiji.analyzeSkeleton.AnalyzeSkeleton_
  */
 
-@Plugin(type = Command.class, menuPath = "Plugins>BoneJ>Inter-trabecular Angles")
+@Plugin(type = Command.class, menuPath = "Plugins>BoneJ>Intertrabecular angles")
 public class IntertrabecularAngleWrapper extends ContextCommand {
```
alessandrofelder: In the original publication the name is "Inter-trabecular Angle" (singular, spelt with a dash) - maybe should be kept that way for consistency? I don't mind capitalization, whatever other plugin menu paths do.

IntegrationTestLogs/IntertrabecularAngle.txt
```diff
@@ -5,7 +5,7 @@ Case 1
 Steps
 -----------------------------------------------------------------------
 1. Open the sample: File > Open Samples > Bat Cochlea Volume
-2. Run plugin BoneJ > Inter-trabecular Angle
+2. Run Plugins > BoneJ > Intertrabecular angles
 3. Slide "Maximum valence" to 5
```
alessandrofelder: see previous comment about spelling of ITA

rimadoma:
Typos fixed, I'll try to spell it "Inter-trabecular" in all the docs as well.

PR #139, opened 2018-09-17
Add fire LUT

rimadoma:
Adds a method that creates a LUT that has the same colour scheme as the _"Fire"_ LUT in ImageJ1.
Thanks @mdoube!


ctrueden:
Do you know that `net.imagej.display.ColorTables` [already has an accessible Fire LUT](https://github.com/imagej/imagej-common/blob/imagej-common-0.27.0/src/main/java/net/imagej/display/ColorTables.java#L48)?

PR #138, opened 2018-09-13
Phone home

mdoube:
This is a bit rough around the edges at the moment, but works well enough. Comments on design improvements are welcome.

PR #140, opened 2018-09-21
Phone home

rimadoma:
Adds BoneJ1 style `UsageReporter`, which allows us to gather data about the popularity of different BonejJ plug-ins if users opt-in. This PR is only different from #138 in that it adds fixes, so that we can add the functionality to a release ASAP. @ctrueden is right that we should probably use `UsageService`, but right now time is short, and I reckon it'll take me less time to fix this PR than to figure out `UsageService`. The downside is that this creates additional work later on.

This PR was merged with more lax standards than usually due time constraints. Affected classes still need more tests, and cleaning up.


mdoube:
Thanks Richard, this is looking much better and working well.

rimadoma:
@mdoube Welp, I did forget about the help-button that doesn't do anything in URO

mdoube:
Ok, quick fix and point release?

Have a nice weekend,

Michael

Sent from ProtonMail mobile

-------- Original Message --------

rimadoma:
@mdoube what would you prefer? No help button or a button despite the concerns I mentioned in your PR?

rimadoma:
@mdoube and what's the specific URL under bonej.org?

mdoube:
Help button that points to bone.org/stats

The info there is correct enough to not be misleading.

Sent from ProtonMail mobile

-------- Original Message --------

rimadoma:
Help button patch released.

PR #147, opened 2018-10-04
Particles: reverse code changes from 10fc038 that introduced an AIOOB

### LINKED ISSUE #146: Particle analyser failed for some particular images
When I ran the Particle analyser(3D), I notice that the plug-in failed to go on, and a bug window jump out as follows:

(Fiji Is Just) ImageJ 2.0.0-rc-68/1.52g; Java 1.8.0_172 [64-bit]; Windows 10 10.0; 783MB of 15251MB (5%)
 
java.lang.ArrayIndexOutOfBoundsException: 1402
	at org.bonej.plugins.ParticleCounter.mergeDuplicates(ParticleCounter.java:2137)
	at org.bonej.plugins.ParticleCounter.joinMappedStructures(ParticleCounter.java:2043)
	at org.bonej.plugins.ParticleCounter.getParticles(ParticleCounter.java:1744)
	at org.bonej.plugins.ParticleCounter.getParticles(ParticleCounter.java:1668)
	at org.bonej.plugins.ParticleCounter.run(ParticleCounter.java:258)
	at ij.IJ.runUserPlugIn(IJ.java:228)
	at ij.IJ.runPlugIn(IJ.java:192)
	at ij.Executer.runCommand(Executer.java:137)
	at ij.Executer.run(Executer.java:66)
	at java.lang.Thread.run(Thread.java:748)

I'm using the newest version of BoneJ as I update it in the Help-update.

I have upload the CT image stacks for your test. As the former 202 slices works well in Particle analyser(3D), when I start add the No. 203 onto the stack, the Particle analyser(3D) corrupted.  
https://1drv.ms/f/s!At8hOBAhBGnMiAeZLoX0JtBNAOIo
Please help me to test it. Many thanks!

mdoube:
The AIOOB only occurred on some images with special conditions,
presumably relating to the particular way the particle labels were
handled in unusual cases.

Fixes #146

### DIFF
--- a/Legacy/bonej/src/main/java/org/bonej/plugins/ParticleCounter.java
+++ b/Legacy/bonej/src/main/java/org/bonej/plugins/ParticleCounter.java
@@ -34,7 +34,6 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
-import java.util.Set;
 import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -552,13 +551,17 @@ public class ParticleCounter implements PlugIn, DialogListener {
 	 * @param lut a look-up table.
 	 * @return total number of duplicate values found.
 	 */
-	private static long countDuplicates(final int[] counter,
-		final List<HashSet<Integer>> map, final int[] lut)
+	private static int countDuplicates(int[] counter,
+		final ArrayList<HashSet<Integer>> map, final int[] lut)
 	{
+		//reset to 0 the counter array
+		final int l = counter.length;
+		counter = new int[l];
+		HashSet<Integer> set = null;
 		for (int i = 1; i < map.size(); i++) {
-			final Set<Integer> set = map.get(i);
+			set = map.get(i);
 			for (final Integer val : set) {
-				final int v = val;
+				final int v = val.intValue();
 				// every time a value is seen, log it
 				counter[v]++;
 				// update its LUT value if value was
@@ -570,7 +573,11 @@ public class ParticleCounter implements PlugIn, DialogListener {
 		minimiseLutArray(lut);
 		// all values should be seen only once,
 		// count how many >1s there are.
-		return Arrays.stream(counter).filter(i -> i > 1).count();
+		int count = 0;
+		for (int i = 1; i < l; i++) {
+			if (counter[i] > 1) count++;
+		}
+		return count;
 	}
 
 	/**
@@ -1961,7 +1968,7 @@ public class ParticleCounter implements PlugIn, DialogListener {
 		final int h = imp.getHeight();
 		final int d = imp.getImageStackSize();
 
-		final List<HashSet<Integer>> map = new ArrayList<>(nParticles + 1);
+		final ArrayList<HashSet<Integer>> map = new ArrayList<>(nParticles + 1);
 
 		final int[] lut = new int[nParticles + 1];
 		// set each label to be its own root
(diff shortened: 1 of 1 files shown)


mdoube:
@alessandrofelder let's get a patch released with this in it so our users can continue their work

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleCounter.java
```diff
@@ -552,13 +551,17 @@ else if (workArray[z][arrayIndex] == FORE &&
...
 	 * @return total number of duplicate values found.
 	 */
-	private static long countDuplicates(final int[] counter,
-		final List<HashSet<Integer>> map, final int[] lut)
+	private static int countDuplicates(int[] counter,
+		final ArrayList<HashSet<Integer>> map, final int[] lut)
 	{
+		//reset to 0 the counter array
+		final int l = counter.length;
+		counter = new int[l];
```
alessandrofelder: Is this clearing of the counter array what prevents the AIOOB from happening in the "special cases" you mention?

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleCounter.java
```diff
@@ -570,7 +573,11 @@ private static long countDuplicates(final int[] counter,
 		minimiseLutArray(lut);
 		// all values should be seen only once,
 		// count how many >1s there are.
-		return Arrays.stream(counter).filter(i -> i > 1).count();
+		int count = 0;
+		for (int i = 1; i < l; i++) {
+			if (counter[i] > 1) count++;
+		}
+		return count;
```
alessandrofelder: I fail to see how the for loop does anything different from the stream variant? Maybe my ignorance.

mdoube: Not sure as I'm still learning about streams. So, maybe my ignorance.

My main concern about it is that the for loop has i, l and 1, which I think might have been mixed up in the stream implementation. For me the for loop is much easier to read and understand.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleCounter.java
mdoube: Maybe. There was also somewhere else an instantiation that got moved inside a while(){}, which might have been responsible.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleCounter.java
mdoube: I think there may have been a couple of things going on in the commit that broke it, and haven't much interest in pursuing it beyond putting it back how it was before when it was working.

alessandrofelder: OK, fair enough

kld123509945: Hey guys, may I asked that if I remove and re-install the BoneJ now, would my CT images work? Well I achieve these CT images from multi-phase flow in the glass beads, and I'm not sure about what's going on with my data. I have many stacks which work well with BoneJ, while only in this team of experiment with this stack and its follow-up experimental images failed....

kld123509945: So exciting about new releases for short tests. Thanks so much for your help!

rimadoma: @alessandrofelder You can do Arrays.stream(arr).skip(1)... to start from 1. But anyhow this was carelessness on my part.

PR #154, opened 2018-11-15
Replace deprecated option with imagej.deleteOtherVersions

mdoube:
Maven is complaining (warning) that `delete.other.versions` has been deprecated and should be replaced with `imagej.deleteOtherVersions`

### DIFF
--- a/IJinstall.sh
+++ b/IJinstall.sh
@@ -10,5 +10,5 @@ if [ "$1" = "" ]
   then
     echo "No ImageJ directory specified"
 else
-    mvn -Dimagej.app.directory="$1" -Ddelete.other.versions=true clean install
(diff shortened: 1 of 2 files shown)


mdoube:
Unfortunately at the moment this build script is throwing lots of annoying errors, despite `mvn clean install` working perfectly and all tests passing:

mdoube:
Not working - this whole script needs a rewrite

mdoube:
This was meant to fix issue #156 but doesn't.

ctrueden:
Once a new pom-scijava is released that includes scijava/pom-scijava@662651b7fafff286d5d5703faf71e6942e33661f, the build will use the scijava-maven-plugin instead, and the appropriate property will be `-Dscijava.deleteOtherVersions=older` although this will be the default behavior. So probably this property should simply be removed from the install scripts.

Issue #164, opened 2019-01-05
ROIs are not properly handled


mdoube:
BoneJ2 cannot handle irregular ROIs for measurements such as BV/TV and Tb.Th.

BoneJ1 could limit measurements to 3D ROIs (series of 2D ROIs) via the ROI manager.

ROI handling is a core functionality and needs to be included.

ctrueden:
@mdoube In case it wasn't already clear from the new ROI classes: the idea is that you can bind a ROI to an image to create another `IterableInterval` that only iterates the samples within the ROI. In this way, you can write an algorithm to process any arbitrary/irregular ROI without needing to complexify the algorithm's API (e.g. to accept an explicit mask object or any such).

It's a little tricky if you need `RandomAccessibleInterval` to e.g. examine neighborhoods, due to the typing: the `Regions.sample` method returns `IterableInterval` even though the thing being sampled is `RandomAccessible`—so the abstraction loses capabilities here.

Maybe @tpietzsch has some thoughts on the best general API strategy here? I think it would be nice to be able to have one object that is `RandomAccessible` as well as `Iterable` for algorithms that want to process irregular regions including the ability to examine the entire sample space as needed.

PR #165, opened 2019-01-22
Remove broken ROI handling behaviour

mdoube:
Modern Thickness was cropping to irregular ROIs rather than masking, which is quite different behaviour than BoneJ1, which might confuse users and lead to wrong results. A rethink of ROI handling is required - hopefully there is a better approach to ROIs in ImageJ2 by now. This code excision is a first step in addressing #164


mdoube:
@rimadoma there are possibly other files to edit such as instructions for testing or documentation - please point me in the right direction.

rimadoma:
@mdoube There's the technical documentation at the wiki in the repository here and then user documentation at imagej.Net

mdoube:
OK @rimadoma docs updated (very good docs BTW - thanks!)

ctrueden:
> Would be good to know how ROIs work in IJ2, but haven't found the time to figure it out yet.

https://imagej.net/ROIs, if you didn't already see it, @alessandrofelder.

We still need to make some Jupyter notebooks illustrating the latest ImgLib2 ROI API.

PR #166, opened 2019-02-08
add outOfBounds utility

alessandrofelder:
simple outOfBounds utility class that avoids code duplication across BoneJ2


alessandrofelder:
Thanks, @mdoube and @ctrueden!

PR #172, opened 2019-04-15
Travis: Correct email address for mdoube

mdoube:


mdoube:
Github wants us to keep merging from master, closing to avoid a messy history. See #183

PR #174, opened 2019-04-17
[maven-release-plugin] prepare release bonej-6.1.1

alessandrofelder:


alessandrofelder:
@mdoube I made a mess of this again - no need to do anything - will try to sort tomorrow - my brain is fried :D

alessandrofelder:
Commit history is awful and cumbersome, but I think this is workable at least. I propose we leave it as is and move on?

Issue #108, opened 2018-08-03
Connectivity is very slow


mdoube:
Processing a 512×512×512 image on my 40-core workstation takes about 30 seconds instead of ~1 sec that the BoneJ1 version takes. BoneJ2 version needs multithreading - currently runs on one core.

rimadoma:
Calling the algorithm through the Op framework creates some overhead, but certainly not the 30-fold increase seen here. It would be interesting to see if a larger image has similar magnitude of speed difference, or whether it evens out.

The parallellization of the code would have to go [here](https://github.com/imagej/imagej-ops/blob/master/src/main/java/net/imagej/ops/topology/eulerCharacteristic/EulerCharacteristic26NFloating.java). I think it would be simpler to run the code inside the op in parallel rather than write the op so that several instances of it can be run at the same time. 

Connectivity is a memory bound problem, i.e. lots of quick accesses to shared RAM. My prediction is that you'll see gains from 2 to n (n << 40) cores, and then performance worsens again after n + 1 cores due parallellization overhead. 

Two parallellization strategies come to mind: either share a [RandomAccess](http://javadoc.scijava.org/ImgLib2/net/imglib2/RandomAccess.html), or process the image stack in sub-regions i.e. chunks. The former might be slow due having to synchronize access to the [RandomAccess], and the latter due cache misses.

Most other plug-ins are slower than in BoneJ1 as well. For me speed is not a top priority: it's more important that both plug-ins _A_ and _B_ work than _A_ works really fast. Also I feel that improving performance is somewhat premature on top of an API that is still changing. A project like this needs what I like to call _Kalashnikov code_: not the most sophisticated, but easy to maintain by anyone even in tough conditions.

mdoube:
Experience shows that synchronising access is usually not the way to go - although my experience is limited to Java 6. Multithreading read-only access to ImageStack was straightforward because each slice lives in a separate array, so all you had to do was point each thread at its own slice. There was a natural chunking strategy. In IJ2, that might not be so clear, although if one is able to restrict an Op thread to work on only a single value of one of the dimensions, then you could easily avoid hitting values twice. This is a read-only process so there's no need to worry about values getting written and read out of sequence - main thing to do is allow the data to flow fast from memory to CPU.

While I agree that it's acceptable to have say a 10% performance hit due to more robust engineering, the current state of Connectivity is ~30× slower, making it a bit tedious for users who are used to the old code.

rimadoma:
[...]

PR #175, opened 2019-04-24
Speed up isBinary() check

mdoube:
Remove high level things from inner loop like Object creation and
HashSet.add(), replacing them with low-level comparison logic.

If needed the same could be done using Object.equals() or similar, to
make isBinary() more generic.

Partially addresses #108 (see also https://github.com/imagej/imagej-ops/commit/dd5a3e3015e9f72e936447ba95c0cac84e822a4d#commitcomment-33284909)

### DIFF
--- a/Modern/utilities/src/main/java/org/bonej/utilities/ElementUtil.java
+++ b/Modern/utilities/src/main/java/org/bonej/utilities/ElementUtil.java
@@ -96,16 +96,13 @@ public final class ElementUtil {
 
 	/**
 	 * Checks whether the interval contains only two distinct values.
-	 * <p>
-	 * NB a hacky brute force approach.
-	 * </p>
 	 *
 	 * @param interval an iterable interval.
 	 * @param <T> type of the elements in the interval.
 	 * @return true if only two distinct values, false if interval is empty
-	 *         or has more colors.
+	 *         or has more values.
 	 */
-	public static <T extends RealType<T> & NativeType<T>> boolean isColorsBinary(
+	public static <T extends RealType<T> & NativeType<T>> boolean isBinary(
 		final IterableInterval<T> interval)
 	{
 		if (interval.size() == 0) {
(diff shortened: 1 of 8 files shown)


Modern/utilities/src/main/java/org/bonej/utilities/ElementUtil.java
```diff
@@ -119,16 +116,33 @@ private ElementUtil() {}
...
-			final T e = cursor.get();
-			colors.add(e.getRealDouble());
-			if (colors.size() > 2) {
-				return false;
+			c = cursor.get().getRealDouble();
+			//if we encounter a different pixel value
+			// than a, assign it to b, and stop
+			if (c != a) {
+				b = c;
+				break;
```
rimadoma: Could have just one loop with

```java
double colors = 1;
while (cursor.hasNext()) {
    c = cursor.get().getRealDouble();
    if (c == a) {
        continue;
    }
    if (colors == 1) {
         b = c;
         colors++;
    } else if (c != b) {
        return false;
    }
}
```

mdoube: true, but it means that for many if not most of the pixels you have to do an extra if, assignment and increment, unnecessarily.

Modern/utilities/src/test/java/org/bonej/utilities/ElementUtilTest.java
```diff
@@ -175,14 +175,14 @@ public void testIsColorsBinary() {
...
 		assertTrue("A BitType image should be binary", isBinary);
 	}
 
 	@Test(expected = NullPointerException.class)
 	public void testIsColorsBinaryThrowsNPEIfIntervalNull() {
-		final boolean result = ElementUtil.isColorsBinary(null);
+		final boolean result = ElementUtil.isBinary(null);
 
 		assertFalse("A null interval should not be binary color", result);
 	}
```
alessandrofelder: I approve the change in method name for clarity and brevity.
In line with the change in method name, the assertion string should lose the "color"?

Modern/utilities/src/test/java/org/bonej/utilities/ElementUtilTest.java
```diff
@@ -195,7 +195,7 @@ public void testIsColorsBinaryReturnsFalseForMulticolor() {
 			.iterator();
 		interval.cursor().forEachRemaining(e -> e.setReal(intIterator.next()));
 
-		final boolean result = ElementUtil.isColorsBinary(interval);
+		final boolean result = ElementUtil.isBinary(interval);
 
 		assertFalse(
 			"An image with more than two colours should not be binary color", result);
```
alessandrofelder: see above

Modern/utilities/src/test/java/org/bonej/utilities/ElementUtilTest.java
```diff
@@ -205,7 +205,7 @@ public void testIsColorsBinaryReturnsFalseForMulticolor() {
 	public void testIsColorsBinaryReturnsFalseIfIntervalEmpty() {
 		final IterableInterval<DoubleType> interval = ArrayImgs.doubles(0);
 
-		final boolean result = ElementUtil.isColorsBinary(interval);
+		final boolean result = ElementUtil.isBinary(interval);
 
 		assertFalse("An empty image should not be binary color", result);
```
alessandrofelder: see above

Modern/utilities/src/test/java/org/bonej/utilities/ElementUtilTest.java
```diff
@@ -214,7 +214,7 @@ public void testIsColorsBinaryReturnsFalseIfIntervalEmpty() {
 	public void testIsColorsBinaryReturnsTrueForMonochrome() {
 		final IterableInterval<DoubleType> interval = ArrayImgs.doubles(2, 2);
 
-		final boolean result = ElementUtil.isColorsBinary(interval);
+		final boolean result = ElementUtil.isBinary(interval);
 
 		assertTrue("Monochrome image should be binary color", result);
 	}
```
alessandrofelder: see above

mdoube: Good catch. I've removed references to colour from the tests.

PR #185, opened 2019-05-09
update scijava parent in pom

alessandrofelder:
Travis build still fails after merging #177 - @imagejan on gitter thinks updating scijava parent to `26.0.0` should fix this - so let's try. I would be interested in what happens underneath if anyone can explain it? Why is Travis on Github happy with the changes, but Travis itself not? Thanks for all your help

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.scijava</groupId>
         <artifactId>pom-scijava</artifactId>
-        <version>24.0.0</version>
+        <version>26.0.0</version>
         <relativePath />
     </parent>
 


imagejan:
Travis, when running on a PR, only _builds the artifact locally_, but doesn't deploy. When a commit to the `master` branch (e.g. a merged PR) is built, the [`build.sh`](https://github.com/scijava/scijava-scripts/blob/master/travis-build.sh) script _builds and deploys a `SNAPSHOT`_ to the maven repository.

Since `maven.imagej.net` moved to `maven.scijava.org`, all builds trying to deploy to the former will fail with a `301` error. That's why both the location of the repository and the goal `deploy-to-scijava` have to be updated. Curtis is filing PRs en masse these days to update affected repositories.

alessandrofelder:
Thanks, everyone! Seems to be fixed now :)

PR #190, opened 2019-05-28
Make pixel mask option not persist. Resets to true on each run.

mdoube:
Not masking leads to bias in the measurements, so it is unsafe to allow
masking to be disabled permanently.


mdoube:
> All good. Any updates to technical or user documentation needed?

Yes - done:
https://imagej.net/BoneJ2#Thickness
https://github.com/bonej-org/BoneJ2/wiki/ThicknessWrapper

alessandrofelder:
Awesome!

PR #197, opened 2020-02-07
Refactor Anisotropy

rimadoma:
@mdoube I had an idea how to improve Anisotropy, and wanted to open this PR for discussion, before it's ready.

The current version blindly shoots sampling vectors from a plane through through the image. This plane needs to be bigger than the stack, so that the vectors have the chance to cover the entire stack, regardless of the plane's orientation. It needs to be _d x d_ to be exact, where _d_ is the largest diagonal of the stack. This means that some of the random vectors are going to miss the stack entirely. The result is that the total length of the MIL vector sampled through the stack will change between runs. I think this is why you see variance in the anisotropy of an empty stack, and why it's not _0.0_.

Here I simply keep generating new random vectors, until a predetermined MIL vector total length is reached. Thus MIL vectors have the same base length (before dividing by interceptions). Attached [here](https://github.com/bonej-org/BoneJ2/files/4172066/ani-comp.txt)
 are some initial results. I think with the new method the anisotropy of an empty stack is close enough to _0.0_ to be attributed to the usual floating point rounding malarkey.

### DIFF
--- a/Modern/ops/src/main/java/org/bonej/ops/mil/ParallelLineGenerator.java
+++ b/Modern/ops/src/main/java/org/bonej/ops/mil/ParallelLineGenerator.java
@@ -1,0 +1,25 @@
+package org.bonej.ops.mil;
+
+import org.joml.Vector3dc;
+
+/**
+ * An interface for random parallel line generators.
+ */
+public interface ParallelLineGenerator {
+    Line nextLine();
+
+    Vector3dc getDirection();
+
+    /**
+     * A line defined as a direction and point it passes through.
+     */
+    final class Line {
+        final Vector3dc point;
+        final Vector3dc direction;
+
+        public Line(Vector3dc point, Vector3dc direction) {
+            this.point = point;
+            this.direction = direction;
+        }
+    }
+}
--- a/Modern/ops/src/main/java/org/bonej/ops/mil/ParallelLineMIL.java
+++ b/Modern/ops/src/main/java/org/bonej/ops/mil/ParallelLineMIL.java
@@ -23,34 +23,27 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 package org.bonej.ops.mil;
 
-import java.util.Objects;
+import java.util.Arrays;
 import java.util.Random;
-import java.util.stream.IntStream;
-import java.util.stream.Stream;
 
 import net.imagej.ops.Contingent;
 import net.imagej.ops.Op;
-import net.imagej.ops.linalg.rotate.Rotate3d;
 import net.imagej.ops.special.function.AbstractBinaryFunctionOp;
-import net.imagej.ops.special.hybrid.BinaryHybridCFI1;
-import net.imagej.ops.special.hybrid.Hybrids;
 import net.imglib2.Interval;
 import net.imglib2.RandomAccess;
 import net.imglib2.RandomAccessible;
 import net.imglib2.RandomAccessibleInterval;
 import net.imglib2.type.BooleanType;
-import net.imglib2.type.numeric.integer.LongType;
-import net.imglib2.type.numeric.real.DoubleType;
 import net.imglib2.util.ValuePair;
 
 import org.joml.Intersectiond;
-import org.joml.Quaterniond;
-import org.joml.Quaterniondc;
 import org.joml.Vector2d;
 import org.joml.Vector3d;
 import org.joml.Vector3dc;
 import org.scijava.plugin.Parameter;
 import org.scijava.plugin.Plugin;
+
+import static org.bonej.ops.mil.ParallelLineGenerator.Line;
 
 /**
  * An op that finds the mean intercept length (MIL) vector of an interval.
@@ -86,30 +79,22 @@ import org.scijava.plugin.Plugin;
  * @author Richard Domander
  */
 @Plugin(type = Op.class)
-public class MILPlane<B extends BooleanType<B>> extends
-	AbstractBinaryFunctionOp<RandomAccessibleInterval<B>, Quaterniondc, Vector3d>
+public class ParallelLineMIL<B extends BooleanType<B>> extends
+	AbstractBinaryFunctionOp<RandomAccessibleInterval<B>, ParallelLineGenerator, Vector3d>
 	implements Contingent
 {
-
-	private static BinaryHybridCFI1<Vector3d, Quaterniondc, Vector3d> rotateOp;
-	private final Random random = new Random();
 	/**
-	 * Number of sampling lines generated per dimension.
+	 * Length of the MIL vector sampled.
 	 * <p>
-	 * Each bin represents a sub-region of the plane where a sampling line
-	 * originates. For example, if bins = 2 then we get four lines originating
-	 * from the four quadrants of the plane.
+	 * The op generates and samples random lines through the interval until their total length
+	 * reaches the given value.
 	 * </p>
 	 * <p>
-	 * If left null, bins is set to <em>d</em>, where <em>d</em> is the largest
-	 * diagonal of the input interval.
-	 * </p>
-	 * <p>
-	 * The higher the number, the more likely you get an accurate result.
+	 * If left null, the length is 100.0 * d, where d = the size of the biggest dimension of the interval.
 	 * </p>
 	 */
 	@Parameter(required = false, persist = false)
-	private Long bins;
+	private Double milLength;
 	/**
 	 * The scalar step between positions on a sampling line where voxels are read.
 	 * <p>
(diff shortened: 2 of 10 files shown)


rimadoma:
I decided to refactor the code so that the generation of random lines is (entirely) separate from sampling the phase changes. I realised that the tests we have are brittle because there's so much randomness involved. With the new structure we can test the phase change sampling with `milOp(img, myTestGenerator)`, where `myTestGenerator` repeats some predetermined lines. This way we know exactly what lines pass through the interval, and what they should encounter.

rimadoma:
Re: the [forum discussion](https://forum.image.sc/t/bonej2-usage-questions/33914/13) I think the total length of the lines sampled for phase changes, and thus the base length of the MIL vector should remain constant for each direction.  I don't think it's a good idea to just continue until something is hit, because that can take a very long time.

Separating between 0, 1 and >1 hits is a problem I was well aware of when originally implementing Anisotropy. I recall @alessandrofelder and I thought at the time that many 0 or 1 hits would mostly happen with "unsuitable" images. With this in mind, maybe the best strategy is to try to develop a clever warning about a potentially unsuitable image? We could also add more data to the output: "% MIL vectors that hit nothing" etc.

rimadoma:
Okay, I did some testing. The findings are:

1. How you count phase changes (back-to-fore only i.e. `current && !previous`, or both i.e. `current != previous`) doesn't really matter
2. MIL vectors that hit nothing are not really a problem
3. Plugin parameters have a huge impact!

First I tested back-to-fore vs. both. Then I tried adding one to intercepts so that even one interception would affect MIL length. I used default parameters, unless otherwise specified. Results were:

Count only back-to-fore
bat-cochlea-volume.tif	0.5121949895513304
xy_sheets.tif	0.9998045952659281 (sampling increment 1.0)

Count both
bat-cochlea-volume.tif	0.5025547625132132
xy_sheets.tif	0.9997861764087104 (sampling increment 1.0)

Count both (and total + 1)
bat-cochlea-volume.tif	0.5029102165798932
xy_sheets.tif	0.9997991405855692 (sampling increment 1.0)

Then I found out that MIL vectors that don't hit anything are rare, unless you have an empty stack. This is because each MIL vector is the sum of all the lines (default = 10 000) of a direction / all the phase changes they encounter. To force the issue I ran _bat-cochlea-volume_  with _2_ lines per dimension (so _4_ per direction) and _1000_ directions. Only _58_ MIL vectors had zero or one interceptions. If you drop lines per dimension and/or directions very low then ellipsoid fitting failing is a bigger issue than MIL vectors that hit nothing anyway.

The third finding is that sampling increment has a big impact on the results. For example, here's the test image [xy_sheets](https://github.com/bonej-org/BoneJ2/files/4215141/xy_sheets.zip) run first with _sampling increment_ _2.3_ and then _1.0_

(increment 2.3)
xy_sheets.tif	0.7773341107759795
xy_sheets.tif	0.8144465990396729
xy_sheets.tif	0.834347062634617
xy_sheets.tif	0.44605003170935276
xy_sheets.tif	0.6540519517056306
(increment 1.0)
xy_sheets.tif	0.9997861764087104
xy_sheets.tif	0.9997846343406255
xy_sheets.tif	0.9997560629338846
xy_sheets.tif	0.9997844830827932
xy_sheets.tif	0.9997724773066062

Of course also dropping lines per dimension and number of directions also increases variance, but not by this much. That also makes more sense, and is less surprising.

@mdoube What's your thoughts? What's the way forward? I suggest we add the constant MIL length and other refactoring introduced in this PR. I don't think we can create a reasonable warning about "bad" images. We'll just have try to document the stochastic nature of the algorithm, and warn about it's sensitivity to input parameters? Users should be aware of "feature size" for a reasonable _sampling increment_?

rimadoma:
I also tested  e6a3528bb12990c1d5b43367c5d220144e51c8f1, with the anisoPPP.tif image from the forum post mentioned above. Now with default parameters its anisotropy is consistently _0.90 - 0.91_, which I think is better.

mdoube:
Interesting findings. Did you look at the point cloud of the rose plot that you get when using different sampling increments? The trouble with using an increment that is <= pixel spacing is that you start to see the anisotropy of the pixel grid itself superimposed on the texture (pixel spacing is greater across the pixels' diagonals than in the orthogonal directions). Because the grid has rotational symmetry, its underlying anisotropy is 0, which may bias anisotropic textures to look less anisotropic than they really are.

Maybe it's a more honest representation of discrete space to allow the pixel grid to shine through than to pretend the pixel grid is not there. The best way to highlight this is to use a test image of binary noise, which should be isotropic, and set the sampling increment to << 1 pixel spacing, then look at the rose plot in the 3D viewer.

Some other ways to try to reduce these kinds of Moire-ish effects is to include some random, sub-pixel jitter in the starting positions. Or even, have variable sampling increments for the same image.

It would be most helpful to include a run-to-stable-result loop into the new Anisotropy, because it is hard to predict exactly which settings will provide a good enough estimate of an image's anisotropy. It would keep sampling until the result doesn't change much.

IIRC xy plates has plates of only 1 pixel thick, which means that a sampling increment of 1 samples the boundaries perfectly, whereas a sampling increment of 2.3 pixels is less efficient at finding boundaries (the ruler is bigger than the feature), so would have to sample more times in more directions to find the texture.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

rimadoma:
>  Did you look at the point cloud of the rose plot that you get when using different sampling increments? The trouble with using an increment that is <= pixel spacing is that you start to see the anisotropy of the pixel grid itself superimposed on the texture (pixel spacing is greater across the pixels' diagonals than in the orthogonal directions).

I didn't, but with this explanation I finally got what you meant by "the pixel grid imposing itself". Cheers! Actually, I think we already talked about his way back when, but again managed not to understand each other, even though we were going on about the same thing.

The problem of a line not necessarily visiting each pixel is passes through is well known in computer graphics, and is explained in [this tutorial](https://lodev.org/cgtutor/raycasting.html#The_Basic_Idea_) I read not long ago. The solution is [Bresenham's algorithmm](https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm) (3D variant [here](https://www.geeksforgeeks.org/bresenhams-algorithm-for-3-d-line-drawing/)). For us it could ensure that the the distance between each voxel sampled is 1, i.e. sampling never jumps diagonally. The flip side is that the algorithm introduces an error to the coordinates sampled, the bigger the longer the line is. Then again this is only a problem when sampling increment is _1_. If it's more, we want sampling to skip voxels, right?

> include some random, sub-pixel jitter in the starting positions

The code's been doing this all along, see [line 260](https://github.com/bonej-org/BoneJ2/blob/e6a3528bb12990c1d5b43367c5d220144e51c8f1/Modern/ops/src/main/java/org/bonej/ops/mil/ParallelLineMIL.java#L260).

> It would be most helpful to include a run-to-stable-result loop into the new Anisotropy,

Unfortunately I have no idea how to achieve this. From what I've seen, even with a large number of directions & lines per direction there's always about 0.015 variance in the result between repeated runs. How would that stabilise? And that's in a good case. Consider how much the results swing in the xy_sheets.tif example above.

mdoube:
> The solution is [Bresenham's algorithmm](https://en.wikipedia.org/wiki/Bresenham%27s_line_algorithm) (3D variant [here](https://www.geeksforgeeks.org/bresenhams-algorithm-for-3-d-line-drawing/)).

Bresenham is a solution to a different problem, which is drawing a line without gaps on a screen using screen dots. We have to draw mathematically perfect vectors in continuous space, then at increments along the line see whether the current pixel is foreground or background. The main choice is whether to assume the pixel location is in the centre of its 'volume' or at the corner, which means you either `Math.floor` or `Math.round`. That is also very efficient, because you just add a fixed (dx, dy, dz) (double precision) increment to the current position, then floor (or round) to get the integer index of the pixel to look up. Bresenham's is way less efficient, which I discovered when trying to find all pixels on an ellipsoid surface.

mdoube:
> > It would be most helpful to include a run-to-stable-result loop into the new Anisotropy,
> 
> Unfortunately I have no idea how to achieve this. From what I've seen, even with a large number of directions & lines per direction there's always about 0.015 variance in the result between repeated runs. How would that stabilise? And that's in a good case. Consider how much the results swing in the xy_sheets.tif example above.

See https://github.com/bonej-org/bonej/blob/810f90299917d8a484102dc1e5cbefee384af8c0/src/org/doube/bonej/Anisotropy.java#L221

All it does is run a minimum number of iterations, calculate DA, then keep running more iterations of randomly selected vectors and intercept lines until the last _n_ DA measurements have a variability that is sufficiently small. The idea is that some textures are sparse or highly variable throughout the image volume, but you don't know how much at the beginning, so the algorithm keeps sampling until the result stabilises. In other words, some images need more dense sampling than others to get a reasonable estimate of their anisotropy. This could be likened to the sampling needed to understand a noisy, irregular signal, versus that required for a regular, clean signal, but not knowing at the outset which one of noisy or clean you are dealing with.

mdoube:
@rimadoma please resolve the conflicts flagged in this PR

PR #199, opened 2020-02-20
Add minimum increment to Anisotropy based on pixel size

rimadoma:
Calculates minimum increment for Anisotropy dynamically based on pixel size, as suggested in PR #197. For now it's just `sqrt(3)` because the plugin doesn't yet handle calibration, but the necessary callbacks are now there for the future.

Also fixes problems in wrapper tests that I noticed when adding new tests for minimum increment. Basically mocking was handled sloppily, mock behaviour wasn't properly reset between tests etc.

### DIFF
--- a/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/AnisotropyWrapper.java
+++ b/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/AnisotropyWrapper.java
@@ -128,7 +128,6 @@ public class AnisotropyWrapper<T extends RealType<T> & NativeType<T>> extends
 	// The default number of lines was found to be sensible after experimenting
 	// with data at hand. Other data may need a different number.
 	private static final int DEFAULT_LINES = 100;
-	private static final double DEFAULT_INCREMENT = 2.3;
 	private static BinaryFunctionOp<RandomAccessibleInterval<BitType>, ParallelLineGenerator, Vector3d> milOp;
 	private static UnaryFunctionOp<Matrix4dc, Optional<Ellipsoid>> quadricToEllipsoidOp;
 	private static UnaryFunctionOp<List<Vector3dc>, Matrix4dc> solveQuadricOp;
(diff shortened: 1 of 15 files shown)


rimadoma:
Before merging I'm still going to add `AbstractWrapperTest` class that does the boiler plate stuff (mock UI, mock UsageReporter etc) shared by most if not all of our wrapper tests. Then the actual test classes can just inherit it, and not repeat the same lines.

Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/AnisotropyWrapper.java
```diff
@@ -302,12 +302,38 @@ private void addResults(final List<Subspace<BitType>> subspaces,
...
+		/* TODO: Get calibration from inputimage.axis(int)
+		   NB: you can't assume that 0, 1, 2 are X, Y, Z axes!
+		   NB: axes can have different units of calibration
+		 */
+		final double px = 1.0;
+		final double py = 1.0;
+		final double pz = 1.0;
+		final double diagonal = px * px + py * py + pz * pz;
+		// Round to 2 decimal places
+		minIncrement = Math.round(Math.sqrt(diagonal) * 100.0) / 100.0;
```
mdoube: This is OK only if uncalibrated distances are used for sampling, i.e. all sampling is based off a unit pixel grid and not pixel spacing in real units.

rimadoma: Hence the comment in the code and in the PR. Anisotropy doesn't handle calibration yet, but that should be a different future PR.

rimadoma:
OK, the PR is ready. JUnit guarentees that a `@BeforeClass` method of a base class is run before child classes, but I read some (old) mentions of a bug in Eclipse. @mdoube Can you double check that the tests run OK when launched from Eclipse?

mdoube:
Tests pass on Eclipse on my system.

PR #200, opened 2020-02-27
New Particle Analyser

mdoube:
Particle Analyser has been refactored to separate concerns. Connected components labelling has been rewritten to be multithreaded and now runs an order of magnitude or two faster than previously, and achieves _O_(n) scaling. The user interface and experience is mostly unchanged, apart from speed.


mdoube:
If someone knows how we can get rid of the merge commit noise, that would be really welcome.

mdoube:
> If someone knows how we can get rid of the merge commit noise, that would be really welcome.

Fixed with a rebase

Legacy/bonej/src/main/java/org/bonej/plugins/ConnectedComponents.java
```diff
@@ -0,0 +1,928 @@
...
+	 * @param map      a map of LUT values.
+	 * @param nbh      a neighbourhood in the image.
+	 * @param centre   current pixel's label (with offset)
+	 * @param IDoffset chunk's ID offset
+	 */
+	private static void addNeighboursToMap(final List<HashSet<Integer>> map, final int[] nbh, final int centre,
+			final int IDoffset) {
+		final int l = nbh.length;
+		int lastNonZero = -1;
+		for (int i = 0; i < l; i++) {
```
rimadoma: Can iterate with `for (final int val : nbh)`, which makes `final int l` unused.

Legacy/bonej/src/main/java/org/bonej/plugins/ConnectedComponents.java
```diff
@@ -0,0 +1,928 @@
...
+		while (ID - IDoffset >= map.size()) {
+			final HashSet<Integer> set = new HashSet<>();
+			set.add(map.size() + IDoffset);
+			map.add(set);
+		}
+	}
+
+	public int getNParticles() {
+		final int n = nParticles;
+		return n;
```
rimadoma: `return nParticles`

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
```diff
@@ -0,0 +1,701 @@
...
+				coOrdinates[i][1] = point.y;
+				coOrdinates[i][2] = point.z;
+				i++;
+			}
+
+			Object[] ellipsoid = null;
+			try {
+				ellipsoid = FitEllipsoid.yuryPetrov(coOrdinates);
+			} catch (final RuntimeException re) {
+
```
rimadoma: In general you should catch only the specific exception types a method might throw. In this case `IllegalArgumentException`. Otherwise we won't be aware that the code is doing something it shouldn't - throwing unexpected exceptions. Might be also worth to log something here, i.e. `IJ.log("Ellipsoid fitting failed")`

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleDisplay.java
```diff
@@ -0,0 +1,477 @@
...
+	}
+
+	/**
+	 * 
+	 * @param imp
+	 * @param ellipsoids
+	 * @param title
+	 * @return ImagePlus containing particles drawn as best-fit solid ellipsoids
+	 */
+	static ImagePlus displayParticleEllipsoids(final ImagePlus imp, final Object[] ellipsoids, final String title) {
```
rimadoma: `title` is always `"Ellipsoids"` so the parameter isn't really needed

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleDisplay.java
```diff
@@ -0,0 +1,477 @@
...
+				}
+				// Add the mesh
+				try {
+					univ.addTriangleMesh(surfacePoint, pColour, "Surface " + p).setLocked(true);
+				} catch (final NullPointerException npe) {
+					IJ.log("3D Viewer was closed before rendering completed.");
+					return;
+				}
+			}
+			p++;
```
rimadoma: Could be more readable with
```
		for (int i = 1; i < surfacePoints.size(); i++) {
			IJ.showStatus("Rendering surfaces...");
			IJ.showProgress(i - 1, nSurfaces);
			final List<Point3f> surfacePoint = surfacePoints.get(i);
			if (surfacePoint.isEmpty()) {
				continue;
			}
			final Color3f colour = getColour(i, colourMode, splitValue, volumes, eigens);
			if (!addMesh(univ, surfacePoint, colour, i)) {
				return;
			}
		}
```
For this you'd need to change the parameter type from `Collection` to `List`

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleDisplay.java
```diff
@@ -0,0 +1,477 @@
...
+
+		final float hue = (float) (angle / Math.PI);
+		final float saturation = 1.0f;
+		final float brightness = 1.0f;
+
+		final int rgb = Color.HSBtoRGB(hue, saturation, brightness);
+		final Color color = new Color(rgb);
+		float red = (float) (color.getRed() / 255d);
+		float green = (float) (color.getGreen() / 255d);
+		float blue = (float) (color.getBlue() / 255d);
```
rimadoma:
```
		float red = color.getRed() / 255.0f;
		float green = color.getGreen() / 255.0f;
		float blue = color.getBlue() / 255.0f;
```

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleDisplay.java
```diff
@@ -0,0 +1,477 @@
...
+		return true;
+	}
+
+	private static int clamp(int value, int min, int max) {
+		if (value < min)
+			return min;
+		if (value > max)
+			return max;
+		return value;
+	}
```
rimadoma: `min` is always `0`, so
```
	private static int clamp(int value, int max) {
		if (value < 0)
			return 0;
		return Math.min(value, max);
	}
```

Legacy/util/src/main/java/org/bonej/geometry/Vectors.java
```diff
@@ -79,4 +81,30 @@ private Vectors() {}
...
+		final double z1 = point1.z - point0.z;
+		final double x2 = point2.x - point0.x;
+		final double y2 = point2.y - point0.y;
+		final double z2 = point2.z - point0.z;
+		final Point3f crossVector = new Point3f();
+		crossVector.x = (float) (y1 * z2 - z1 * y2);
+		crossVector.y = (float) (z1 * x2 - x1 * z2);
+		crossVector.z = (float) (x1 * y2 - y1 * x2);
+		return crossVector;
+	}
```
rimadoma: I know `org.scijava.vecmath` is still needed for 3DViewer and old marching cubes, but since ImageJ is not going to support it forever, and BoneJ has already started to use `org.joml.joml` for linear algebra, we should try to avoid adding code that uses the artefact. 

In some ways it would be better to convert the `Point3f` to `org.joml.Vector3f` and call the perfectly good `crossProduct()` there, but I realise that'd be a bit inefficient. 

Since nothing else than `ParticleAnalysis` needs the method, lets make it a `private static` member of that class, so that we can do away with the `Vectors` utility class asap.

Legacy/bonej/src/main/java/org/bonej/plugins/ConnectedComponents.java
```diff
@@ -0,0 +1,928 @@
...
+		//merge neighbour networks and generate a LUT
+		final int[][] lut = generateLut(chunkMaps, chunkIDOffsets);
+		
+		// rewrite the pixel values using the LUT
+		applyLUT(particleLabels, lut, chunkIDOffsets, startSlices, w, h, nSlices);
+
+		return particleLabels;
+	}
+
+	private static int[][] generateLut(ArrayList<ArrayList<HashSet<Integer>>> chunkMaps, int[] chunkIDOffsets) {
```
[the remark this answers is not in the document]
mdoube: Agreed - Here the specific types are used purposefully.

Legacy/bonej/src/main/java/org/bonej/plugins/ConnectedComponents.java
```diff
@@ -0,0 +1,928 @@
...
+	 * collisions. Transitively connected labels are joined when they are added to the same, 
+	 * lower-indexed, HashSet.
+	 * <br/>
+	 * By iterating downwards from the highest index, this algorithm resembles buckets
+	 * splashing water from the uppermost to the lowermost.
+	 * It is named after <i>Bucket Fountain</i>, a kinetic sculpture by Burren and Keen (1969),
+	 * installed in Cuba Mall, Wellington, New Zealand.
+	 * 
+	 * @see <a href="https://en.wikipedia.org/wiki/Bucket_Fountain">Wikipedia: Bucket Fountain</a>
+	 * @param chunkMaps
```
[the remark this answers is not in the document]
mdoube: OK, noted, will fix in a later style PR. Made my IDE more fussy about Javadoc.

Legacy/bonej/src/main/java/org/bonej/plugins/ConnectedComponents.java
```diff
@@ -0,0 +1,928 @@
...
+					if (minLabel < IDoffset) {
+						priorMap.get(minLabel - priorIDoffset).addAll(set);
+						set.clear();
+						continue;
+					}
+					// move whole set's contents to a lower position in the map
+					if (minLabel < i + IDoffset) {
+						map.get(minLabel - IDoffset).addAll(set);
+						set.clear();
+						continue;
```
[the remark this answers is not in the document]
mdoube: OK

Legacy/bonej/src/main/java/org/bonej/plugins/ConnectedComponents.java
```diff
@@ -0,0 +1,928 @@
...
+				}
+			}
+		}
+
+		// check the hashMap for duplicate appearances and merge sets downwards
+		boolean somethingChanged = true;
+		while (somethingChanged) {
+			somethingChanged = false;
+			Iterator<Map.Entry<Integer, HashSet<Integer>>> it = hashMap.entrySet().iterator();
+			while (it.hasNext()) {
```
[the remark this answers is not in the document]
mdoube: Thanks. Was concerned about concurrent modification, but for this loop, which doesn't create or destroy entries, just modifies them, the shorthand syntax without an Iterator is cleaner.
https://stackoverflow.com/questions/1066589/iterate-through-a-hashmap

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
```diff
@@ -0,0 +1,701 @@
...
+				final Point3f point = pointIter.next();
+				coOrdinates[i][0] = point.x;
+				coOrdinates[i][1] = point.y;
+				coOrdinates[i][2] = point.z;
+				i++;
+			}
+
+			Object[] ellipsoid = null;
+			try {
+				ellipsoid = FitEllipsoid.yuryPetrov(coOrdinates);
```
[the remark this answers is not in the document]
mdoube: A check and exception should probably go in FitEllipsoid.yuryPetrov() so that calling classes are warned that they are not getting an ellipsoid back for the data they sent to the method.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleDisplay.java
mdoube: Fixed, good catch.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleDisplay.java
mdoube: Done.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleDisplay.java
mdoube: You are correct for this particular use case. But as written it is more able to be upstreamed because it can handle any legal int. Optimising for a specific client method will make it brittle for another future client method. What it doesn't do is check to make sure that min < max.

Legacy/util/src/main/java/org/bonej/geometry/Vectors.java
mdoube: OK, good strategy, will do. Can convert the methods as ImageJ moves to SciView and joml. In the meantime as long as the legacy 3DViewer is there and works well and is convenient for users, it's a stable solution.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleDisplay.java
mdoube: True & done.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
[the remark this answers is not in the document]
rimadoma: Have a look at [QuadricToEllipsoid](https://github.com/bonej-org/BoneJ2/blob/master/Modern/ops/src/main/java/org/bonej/ops/ellipsoid/QuadricToEllipsoid.java). I found out that you had to do the eigen decomposition for the parameters that come out of YuryPetrov to find out if what you have is indeed an ellipsoid. 

I can't explain exactly why it wasn't enough to look at the signs or determinant of the parameters in the quadric solution, but in practice that caused "this is not an ellipsoid" way too often.

Incidentally, I'm not sure if it makes sense for `QuadricToEllipsoid` to be an op because it is so specific, so if you find that with little modifications you can make it into a utility that benefit both old and modern code, feel free to do so.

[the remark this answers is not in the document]
mdoube: From my reading it seems sufficient to check that the eigenvalues are all > 0, because the ellipsoid matrix has to be _positive definite_. We have the eigenvalues already to calculate the radii. So just throwing an exception if any of them is <= 0 should be enough:

```
final EigenvalueDecomposition E = (EigenvalueDecomposition) matrices[3];
		final Matrix eVal = E.getD();
		final Matrix diagonal = MatrixUtils.diag(eVal);
		final int nEvals = diagonal.getRowDimension();
		final double[] radii = new double[nEvals];
		for (int i = 0; i < nEvals; i++) {
			final double eigenvalue = diagonal.get(i, 0);
			if (eigenvalue <= 0) {
				throw new IllegalArgumentException("Ellipsoid eigenvalues must be greater than 0 "
						+ "(positive definite matrix)");
			}
			radii[i] = Math.sqrt(1 / eigenvalue);
		}
```
Incidentally, this check prevents a NaN result or Math.sqrt spewing because it prevents calculation of 1/0 and Math.sqrt(negative number).

PR #203, opened 2020-03-03
Code style improvements, after comments on PR #200

mdoube:
- a lot of Javadoc
- exceptions and handling

### DIFF
--- a/Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
+++ b/Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
@@ -31,7 +31,6 @@ import java.util.ListIterator;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.bonej.geometry.FitEllipsoid;
-import org.bonej.geometry.Vectors;
 import org.bonej.util.Multithreader;
 import org.scijava.vecmath.Point3f;
 
@@ -44,10 +43,17 @@ import ij.ImageStack;
 import ij.measure.Calibration;
 import marchingcubes.MCTriangulator;
 
+/**
+ * Perform analysis of particles
+ * @author Michael Doube
+ *
+ */
 public class ParticleAnalysis {
 
+	/** list of particle sizes */
 	private long[] particleSizes;
 
+	/** Constructor */
 	public ParticleAnalysis() {
 
 	}
@@ -64,7 +70,7 @@ public class ParticleAnalysis {
 	 * @param imp            Input image. Needed for calibration
 	 * @param particleLabels Particle label image array
 	 * @param workArray      Binary work array
-	 * @param particleSizes  List of particle sizes in pixels indexed by particle ID
+	 * @param nParticles     Number of particles in the image
 	 * @param phase          Foreground or background
 	 * @param doExclude      true to remove all particles touching a side
 	 * @param min            minimum volume in calibrated units to include
@@ -239,6 +245,7 @@ public class ParticleAnalysis {
 	 *
 	 * @param particleLabels particle label image array
 	 * @param nParticles     number of particles
+	 * @return array of particle sizes
 	 */
 	public long[] getParticleSizes(final int[][] particleLabels, final int nParticles) {
 		final int d = particleLabels.length;
@@ -276,6 +283,13 @@ public class ParticleAnalysis {
 		return this.particleSizes.clone();
 	}
 
+	/**
+	 * Calculate calibrated volumes of the particles
+	 * 
+	 * @param imp ImagePlus, used only for its calibration information
+	 * @param particleSizes list of particle sizes in pixel counts
+	 * @return array of particle sizes in calibrated units
+	 */
 	static double[] getVolumes(final ImagePlus imp, final long[] particleSizes) {
 		final Calibration cal = imp.getCalibration();
 		final double voxelVolume = cal.pixelWidth * cal.pixelHeight * cal.pixelDepth;
@@ -373,6 +387,14 @@ public class ParticleAnalysis {
 		return meanStdDev;
 	}
 
+	/**
+	 * Calculate Eigenvalue decompositions of all the particles
+	 * 
+	 * @param imp ImagePlus, used for calibration
+	 * @param particleLabels label image array
+	 * @param centroids list of particle centroids
+	 * @return list of EigenvalueDecompositions
+	 */
 	static EigenvalueDecomposition[] getEigens(final ImagePlus imp, final int[][] particleLabels,
 			final double[][] centroids) {
 		final Calibration cal = imp.getCalibration();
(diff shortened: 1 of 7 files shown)


Legacy/bonej/src/main/java/org/bonej/plugins/ParticleDisplay.java
```diff
@@ -425,6 +440,40 @@ static void display3DOriginal(final ImagePlus imp, final int resampling, final I
...
+				// red if over
+				colour = new Color3f(1.0f, 0.0f, 0.0f);
+			} else {
+				// yellow if under
+				colour = new Color3f(1.0f, 1.0f, 0.0f);
+			}
+		} else if (colourMode == ORIENTATION) {
+			colour = ParticleDisplay.colourFromEigenVector(eigens[p]);
+		}
+		return colour;
```
rimadoma: In principle, if we knew that colour modes might change, a way to defend against coding errors here would be
```
private enum ColourMode = {GRADIENT, SPLIT, ORIENTATION};
...
else if (colourMode == ORIENTATION) {
  return ParticleDisplay.colourFromEigenVector(eigens[p]);
}

throw new AssertionError("Code should not reach here - not all colour modes are covered");
```

Even if black was a valid default colour, then `DEFAULT` should be one additional enum value and `else if` clause.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleDisplay.java
```diff
@@ -450,13 +499,19 @@ static Color3f colourFromEigenVector(EigenvalueDecomposition eigen) {
...
 		return new Color3f(red, green, blue);
 	}
 
+	/**
+	 * Check whether ellipsoid radii are NaN
+	 * 
+	 * @param radii list of radii
+	 * @return true if all radii are not NaN
+	 */
 	private static boolean isRadiiValid(final double[] radii) {
```
rimadoma: I noticed that in `displayEllipsoids` after you call `isRadiiValid` you check that `a < b < c`. That could be performed inside the method as well.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
```diff
@@ -692,10 +745,38 @@ private static double getSurfaceArea(final List<Point3f> points) {
 			Object[] ellipsoid = null;
 			try {
 				ellipsoid = FitEllipsoid.yuryPetrov(coOrdinates);
-			} catch (final RuntimeException re) {
-
+			} catch (final IllegalArgumentException re) {
+				IJ.log("Could not fit ellipsoid to particle: "+re.getMessage());
+			} catch (final Exception e) {
```
rimadoma: Should only catch `IllegalArgumentException` because that's the only thing we know `yuryPetrov` can throw. If it suddenly starts throwing something else, we're in trouble and should be alerted by seeing the exception come up.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleDisplay.java
```diff
@@ -466,7 +521,18 @@ private static boolean isRadiiValid(final double[] radii) {
...
 
+	/**
+	 * Limit value to within minimum and maxium values
+	 * 
+	 * @param value the input value
+	 * @param min the allowed minimum
+	 * @param max the allowed maximum
+	 * @return value, or min if value < min; max if value > max
+	 */
 	private static int clamp(int value, int min, int max) {
```
rimadoma: I forgot to say in the last PR that you can document exceptions with
```
* @throws IllegalArgumentException if min >  max 
*/
private static int clamp(int value, int min, int max)  throws IllegalArgumentException {
```
Not only is it informative for developers, but your IDE will be clever enough to warn:
```
try {
  clamp(value, min, max);
} catch (Exception e) { // Should catch IllegalArgument here!
```

Legacy/util/src/main/java/org/bonej/geometry/FitEllipsoid.java
```diff
@@ -41,8 +41,10 @@ OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
  */
 public final class FitEllipsoid {
 
+	/** Random number generator*/
 	private static Random rng = new Random();
 
+	/** Constructor */
 	private FitEllipsoid() {}
```
rimadoma: It's not necessary to comment `private` fields or methods. Especially if they're trivial.

Legacy/util/src/test/java/org/bonej/geometry/FitEllipsoidTest.java
```diff
@@ -50,6 +50,27 @@ public void testYuryPetrov() {
 		assertArrayEquals(centre, (double[]) result[0], 1e-2);
 		assertArrayEquals(radii, (double[]) result[1], 1e-2);
 	}
+	
+	@Test(expected = IllegalArgumentException.class)
+	public void testYuryPetrovCylinder() {
```
rimadoma: Good test to see that expected exception is thrown.

Legacy/util/src/test/java/org/bonej/geometry/FitEllipsoidTest.java
```diff
@@ -50,6 +50,27 @@ public void testYuryPetrov() {
...
+		//generate points on a cylinder with d = h, centred on 0, 0, 0.
+		final int r = 50;
+		final int nPoints = 100;
+		final double[][] points = new double[(int)(2 * r * nPoints)][3];
```
rimadoma: All variables in `(int)(2 * r * nPoints)` are `int` so casting is redundant.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
mdoube: There's also a RuntimeException that gets thrown by JAMA if the Matrix is singular, in `LUDecomposition.solve()`:
```java
	/**
	 * Solve A*X = B
	 *
	 * @param B
	 *            A Matrix with as many rows as A and any number of columns.
	 * @return X so that L*U*X = B(piv,:)
	 * @exception IllegalArgumentException
	 *                Matrix row dimensions must agree.
	 * @exception RuntimeException
	 *                Matrix is singular.
	 */

	public Matrix solve(final Matrix B) {
		if (B.getRowDimension() != m) {
			throw new IllegalArgumentException("Matrix row dimensions must agree.");
		}
		if (!this.isNonsingular()) {
			throw new RuntimeException("Matrix is singular.");
		}
```
This might better be caught and handled by `yuryPetrov()`, but still it will get passed up the chain to client code that needs to know that an ellipsoid could not be fitted to the points it submitted to `yuryPetrov()`.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleDisplay.java
mdoube: This error is never logged in practice, I suspect because the eigendecomposition sorts the eigenvalues and correspondingly the radii.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
rimadoma: In that case you can write
```
case (final IllegalArgumentException | RuntimeException e) {
    IJ.log("Could not fit ellipsoid to particle: "+e.getMessage());
}
```

Issue #193, opened 2019-09-04
Slice Geometry throws NullPointerExceptions when calculating Thickness values


alessandrofelder:
**Bug description**
When running Slice Geometry on a stack of binary images, NullPointerExceptions are thrown in some threads.

```
Exception in thread "Thread-29307" java.lang.NullPointerException
	at sc.fiji.localThickness.EDT_S1D.run(EDT_S1D.java:208)
	at sc.fiji.localThickness.LocalThicknessWrapper.processImage(LocalThicknessWrapper.java:127)
	at org.bonej.menuWrappers.LocalThickness.processThicknessSteps(LocalThickness.java:266)
	at org.bonej.menuWrappers.LocalThickness.getLocalThickness(LocalThickness.java:163)
	at org.bonej.plugins.SliceGeometry$SliceThread.run(SliceGeometry.java:1005)
```

**To Reproduce**
With the latest Fiji and BoneJ2 installed, 
1. Open the test image provided [in this forum post](https://forum.image.sc/t/problem-with-bonej2-slice-geometry/29214)
2. Run Plugins > BoneJ > Slice Geometry
3. Tick "Process Stack", and then OK
4. Log shows NullPointerExceptions
5. Results table column Mean3DThickness has lots of zero and NaN entry

**Expected behavior**
No NullPointerExceptions, mean thickness values are >0, finite, and the same for each slice (in this test image)

**Environment**
 - Ubuntu 16.04
 - java 1.8.0_171

mdoube:
This looks like a problem in the reimplementation of Local Thickness and the way it handles 2D slices, or perhaps the way that Local Thickness (complete process) is called slice by slice on a stack by Slice Geometry in BoneJ2. It might be some kind of thread collision, because Thickness itself is multithreaded and so might be the slice processing in Slice Geometry.

This bug does not exist in BoneJ1.

mdoube:
> Results table column Mean3DThickness has lots of zero and NaN entry

This is probably due to a previous run of Slice Geometry creating a 3D thickness column, which is then filled with dummy entries because the 3D option is not selected in a subsequent run. Probably not a symptom of this bug.

rimadoma:
@mdoube Maybe you added something to Thickness in BoneJ1 that allowed it processing 2D (single slice) images? Something that isn't in the original `sc.fiji.LocalThickness_` plugin the implementation now calls. That's where the `NullPointerException` happens at least.

rimadoma:
[...]

PR #212, opened 2020-03-10
Include a specific version for LocalThickness_ (4.0.3)

mdoube:
- as recommended in issue #193

Advice is to put Maven version properties in the top-level POM.


rimadoma:
We can do this, I'm not opposed, but since pom-scijava 29.0.0, which uses 4.0.3 as default, is coming out soon we might as well wait for that? Then we don't have to remember to take out the specific version again.

ctrueden:
:+1: It's great to pin to a specific version to move forward quickly, and then remove the pin again once the next pom-scijava is out and you update to it. Just remember to remove the obsolete pins then. 😄

PR #222, opened 2020-03-24
Fix javadoc warnings

mdoube:
Last touches prior to release


Legacy/bonej/src/main/java/org/bonej/plugins/ConnectedComponents.java
```diff
@@ -68,7 +68,7 @@ public ConnectedComponents() {
 	/**
 	 * Run connected components filter on a binary image
 	 * 
-	 * @param imp   Input ImagePlus, must be 2D or 3D and binary (0 & 255)
+	 * @param imp   Input ImagePlus, must be 2D or 3D and binary (0 or 255)
```
rimadoma: Writing down assumptions about the parameters is good documentation 👍 

Btw, when you do want to use an ampersand, write `&amp;`

PR #224, opened 2020-03-25
Bump to next development cycle

mdoube:
Signed-off-by: Michael Doube <michael@doube.net>

PR #225, opened 2020-03-26
[maven-release-plugin] prepare release bonej-7.0.0

mdoube:
Please merge before #224


rimadoma:
Hang on, why are there `SNAPSHOT` versions in `release.properties`? When was the script run?

mdoube:
> Hang on, why are there `SNAPSHOT` versions in `release.properties`? When was the script run?

The snapshots (7.0.1-SNAPSHOT) appear to relate to some kind of 'dev' designation. It was run on 6.1.2 snapshot, just after PR #223 .

rimadoma:
OK, no worries. I think I figured it out: the `dev` tags just refer to the next development version, which are `SNAPSHOT`

Issue #219, opened 2020-03-23
Include a progress bar indicator while fitting ellipsoids


mdoube:
**Describe the bug**
In ellipsoid factor, it would be helpful for the user to get some idea of progress by a progress bar indicator. This is complicated by ellipsoid fitting being performed within a `ParallelStream`:
 ```java
QuickEllipsoid[] skeletonSeededEllipsoids = skeletonPoints.parallelStream().map(sp -> medialOptimisation.calculate(pixels, sp)).filter(Objects::nonNull).toArray(QuickEllipsoid[]::new);
```
Which makes it difficult to extract the current position within the array of skeleton points on which  `parallelStream()` is working.

[One solution](https://github.com/bonej-org/BoneJ2/pull/209#discussion_r395554530) proposed by @rimadoma is to (if I understand it correctly) drop an `AtomicInteger` into the fitting logic, which is then used by a progress bar updater:
 ```java
final AtomicInteger progress = new AtomicInteger();
...
stream.parallel().peek(x -> {
  statusService.showProgress(progress.getAndIncrement(), max);
}).forEach(x -> ...);
```

**Expected behavior**
The progress bar smoothly updates as ellipsoid fitting continues.

**Context**
EF1 used the `Multithreader` to run in threads and so had an `AtomicInteger` readily available to each thread to update the status bar. (Note the redundant `counter.getAndAdd()`, when `ai.get()` would do).

```java
	final AtomicInteger ai = new AtomicInteger(0);
	final AtomicInteger counter = new AtomicInteger(0);
	final Thread[] threads = Multithreader.newThreads();
	for (int thread = 0; thread < threads.length; thread++) {
		threads[thread] = new Thread(new Runnable() {
			public void run() {
				for (int i = ai.getAndAdd(skipRatio); i < nPoints; i = ai.getAndAdd(skipRatio)) {
						ellipsoids[i] = optimiseEllipsoid(imp, pixels, skeletonPoints[i], unitVectors, i);
					IJ.showProgress(counter.getAndAdd(skipRatio), nPoints);
					IJ.showStatus("Optimising ellipsoids...");
				}
			}
		});
	}
	Multithreader.startAndJoin(threads);
```

mdoube:
Closed with bonej-7.0.1

PR #226, opened 2020-03-30
Address issue #219 and add little fixes

rimadoma:
This PR fixes issue #219, and fixes other small, mostly stylistic things in EllipsoidFactorWrapper

### DIFF
--- a/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/EllipsoidFactorWrapper.java
+++ b/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/EllipsoidFactorWrapper.java
@@ -186,8 +186,8 @@ public class EllipsoidFactorWrapper extends ContextCommand {
 		int counter = 0;
 		for(int i = 0; i<runs; i++) {
 			//optimise ellipsoids
-			final QuickEllipsoid[] ellipsoids = runEllipsoidOptimisation(inputImgPlus);
-			if (ellipsoids.length == 0) {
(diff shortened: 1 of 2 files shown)


rimadoma:
@alessandrofelder Are the conversions in `EllipsoidFactorWrapper` on line 295 necessary?

alessandrofelder:
Some of the images that are divided there consist of integers, so ensuring a float division is important - so I think at least one of the conversions is needed, maybe not both - but don't have time to test that right now

mdoube:
> This PR fixes issue #219, and fixes other small, mostly stylistic things in EllipsoidFactorWrapper

Please check performance implications of the changes and report them (just run before the commits and after the commits on the same image with the same settings and copy paste the stdout output here). We've found that EF is particularly sensitive to small changes.

rimadoma:
```
Old
[INFO] Found 71 ellipsoids in 5734 ms
[INFO] Found maximal ellipsoids in 116 ms

New
[INFO] Found 72 ellipsoids in 5829 ms
[INFO] Found maximal ellipsoids in 192 ms
```

NB These values change a lot from run to run for both old and new

alessandrofelder:
In terms of speed, newer version is slightly faster than current master for me 👍

mdoube:
> In terms of speed, newer version is slightly faster than current master for me +1

We must do this on every PR.

rimadoma:
@alessandrofelder Fixed now. I dropped the "avoid raw types" commits for now.

alessandrofelder:
I am happy with this now :)

PR #228, opened 2020-03-31
Test Anisotropy with a binary noise image

rimadoma:


rimadoma:
> Did you performance test this change?

No, but 
1. There aren't that many calls (not in the main loop)
2. EF uses random gaussian distributed numbers, which are more complex to generate. The Mersenne twister here requires just your usual bit shifting.

PR #230, opened 2020-04-09
Javadoc

mdoube:
Correct an HTML character entity from `&lte` to `&lt` to render the `<` character.


Modern/ops/src/main/java/org/bonej/ops/mil/ParallelLineMIL.java
```diff
@@ -169,7 +169,7 @@ public boolean conforms() {
 	/**
 	 * Sets the seed of the underlying random number generator
 	 * <p>
-	 * Only affects the random offset 0 &lte; o &lte; increment added to the sampling points
+	 * Only affects the random offset 0 &lt; o &lt; increment added to the sampling points
```
rimadoma, line 172: Should be `0 &le; o &lt;`

Issue #236, opened 2020-04-30
Output max feret point coordinates


mdoube:
**Describe the bug**
See this forum post
https://forum.image.sc/t/maximum-feret-diameter-of-irregular-3d-shape/37025/4

mdoube:
Fixed by #237 and #239

PR #237, opened 2020-04-30
Report feret extrema points alongside feret diameter for each particle

mdoube:
Addresses #236

### DIFF
--- a/Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
+++ b/Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
@@ -24,6 +24,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 package org.bonej.plugins;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
@@ -41,7 +42,6 @@ import ij.IJ;
 import ij.ImagePlus;
 import ij.ImageStack;
 import ij.measure.Calibration;
-import ij.plugin.filter.PlugInFilter;
 import marchingcubes.MCTriangulator;
 import sc.fiji.analyzeSkeleton.AnalyzeSkeleton_;
 import sc.fiji.analyzeSkeleton.SkeletonResult;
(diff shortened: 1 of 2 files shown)


Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
```diff
@@ -718,20 +718,25 @@ private static double getSurfaceArea(final List<Point3f> points) {
...
+	 * @return Feret diameters and x, y, z coordinates of the two feret points of each surface,
+	 * packed in a double so that the feret diameter of particle i is found at [i * 7] and the
+	 * points' coordinates are in the following 6 positions in ax, ay, az, bx, by, bz order.
 	 */
 	static double[] getFerets(final List<List<Point3f>> surfacePoints) {
 		Thread[] threads = Multithreader.newThreads();
 		final int nSurfaces = surfacePoints.size();
-		final double[] ferets = new double[nSurfaces];
+		//distance, xa, ya, za, xb, yb, zb
+		final double[] ferets = new double[nSurfaces * 7];
```
rimadoma, line 729: Would be clearer with a `final double[][] ferets = new double[nSurfaces][7];`. This does break API, but that's not a problem since the method is package-private (and only used in one place).

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
```diff
@@ -749,10 +756,21 @@ private static double getSurfaceArea(final List<Point3f> points) {
 						itb = surface.listIterator(ita.nextIndex());
 						while (itb.hasNext()) {
 							b = itb.next();
-							feret = Math.max(feret, a.distance(b));
+							final double distance = a.distance(b);
+							if (distance > feret) {
+								feret = distance;
+								feretA = a;
+								feretB = b;
+							}
```
rimadoma, line 764: Since the list is not mutated, this can be a simple nested loop:
```
double feret = 0;
Point3f feretA = new Point3f();
Point3f feretB = new Point3f();
for (int p = 0; p < surface.size() - 1; p++) {
  final Point3f a = surface.get(p);
  for (int q = p + 1; q < surface.size(); q++) {
    final Point3f b = surface.get(q);
    final double distance = a.distance(b);
    if (distance > feret) {
      feret = distance;
      feretA = a;
      feretB = b;
    }
  }
}
```

rimadoma: Btw, what should happen if there's only one point in the list? `feret = 0`?

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
mdoube: Agreed, thanks.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
mdoube: Good question, but it shouldn't happen because the point lists are output from the surface mesher which is putting points in in groups of 3. The smallest should be a tetrahedron, so 4 triangles and 12 points. Empty surfaces (nPoints = 0) are already handled.

rimadoma: I think that assumption should be documented and checked for:
```
@param surfacePoints points from a surface mesher
@throws IllegalArgumentException if there are less than 12 points (a tetrahedron)
static double[] getFerets(final List<List<Point3f>> surfacePoints) throws IllegalArgumentException {
    if (surfacePoints.size() < 12) {
        throw new IllegalArgumentException("Mesh has too few points");
    }
    ...
}
```

mdoube: > Empty surfaces (nPoints = 0) are already handled.

This exception checking finds a few surfaces with 0 points. Previously only `null` surfaces were handled.

Issue #235, opened 2020-04-29
Make Ellipsoid Factor macro recordable


mdoube:
**Describe the bug**
Ellipsoid Factor (IJ2 style) is not macro recordable. No entry appears in the Macro Recorder when Ellipsoid Factor completes.

**To Reproduce**
Steps to reproduce the behaviour:
1. Run the Macro Recorder
2. Run Ellipsoid Factor

**Expected behavior**
An Ellipsoid Factor command should appear in the Macro Recorder

**Screenshots**
![image](https://user-images.githubusercontent.com/87103/80576338-9a03dc00-8a37-11ea-899a-7c9da8193b3b.png)

**Additional context**
Reported on the forum https://forum.image.sc/t/using-ellipsoid-factor-in-a-macro/36975

It is possible to record `run("Ellipsoid Factor");` but only if you cancel the run from the dialog box. There should be a long string of variables listed in the recorded macro command.

Also, all the other recorded commands are noisy and should be hidden from the user.

Other IJ2 ('Modern') plugins can record, for example [`Anisotropy`](https://github.com/bonej-org/BoneJ2/blob/master/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/AnisotropyWrapper.java):
```
run("Anisotropy", "inputimage=net.imagej.ImgPlus@274042d9 directions=2000 lines=10000 samplingincrement=1.73 recommendedmin=true printradii=true printeigens=true displaymilvectors=false")
```

rimadoma:
I don't think this is a quick fix. Maybe a [hacky workaround](https://forum.image.sc/t/scijava-command-macro-recording/35056/14) is possible, but AFAIK ImageJ2 style commands are intended to be scripted with Python. I know it's a big ask to let go of macros, but my working assumption during the project was always that they were a thing of the past. And I assumed macro-recording would get replaced by something else as well. It was a mistake not to discuss this more though.

mdoube:
> I don't think this is a quick fix.

Is there not a pattern we can copy from `Anisotropy`, which is also a pure IJ2 plugin and is perfectly recorded by the macro recorder?

mdoube:
> ImageJ2 style commands are intended to be scripted with Python

In that case it would for us to give users a hint in the docs about what commands to include. The beauty of the macro recorder is that users don't have to do much and the code is written for them, which is a nice bump to get started with simple procedural programming.

rimadoma:
[...]

PR #238, opened 2020-05-03
Fix issue #235

rimadoma:
Problem was that macro recording recorded the call to Skeletonize within EF. When _seedOnTopologyPreserving = false_ recording worked fine.

### DIFF
--- a/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/EllipsoidFactorWrapper.java
+++ b/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/EllipsoidFactorWrapper.java
@@ -33,6 +33,7 @@ import static org.bonej.wrapperPlugins.wrapperUtils.Common.cancelMacroSafe;
 
 import ij.ImagePlus;
 import ij.ImageStack;
+import ij.plugin.frame.Recorder;
 
 import java.util.ArrayList;
 import java.util.Collection;
@@ -50,7 +51,6 @@ import java.util.stream.Stream;
 
 import net.imagej.axis.DefaultLinearAxis;
 import net.imagej.units.UnitService;
-import net.imglib2.type.NativeType;
 import net.imglib2.type.numeric.integer.ByteType;
 import net.imglib2.type.numeric.integer.IntType;
 import net.imglib2.type.numeric.integer.LongType;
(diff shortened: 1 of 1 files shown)


rimadoma:
@mdoube I haven't figured out yet how to disable the `selectWindow("foo")` lines that result from the output images of EF

rimadoma:
@mdoube Other commands such as ITA might require a similar fix, but let's do them in a separate PR so that this one can go out ASAP

mdoube:
> how to disable the `selectWindow("foo")` lines that result from the output images of EF

These should never really be turned into full images in userland - they have to stay headless. In IJ1 that was achieved by making an `ImagePlus` then calling `.show()` on it for display. Or not even making an `ImagePlus` and leaving the pixel data as an `ImageStack` or raw arrays.

rimadoma:
> These should never really be turned into full images in userland - they have to stay headless. In IJ1 that was achieved by making an ImagePlus then calling .show() on it for display.

I'm wondering if the images need to be output parameters at all. We could just have
```
List<ImgPlus> resultImages = doStuff();
if (!IJ1Helper.isMacro()) {
    resultImages.forEach(o -> uiService::show)
}
```
Then if you wanted to do further stuff to the images, you could still grab them in a macro with selectWindow("Seed points"). Anyhow, I recommend asking the forum about macros and commands.

@alessandrofelder Would Python scripting EF still work if the images were not output parameters?

rimadoma:
> The skeletonisation result flashes up and is then hidden: it should never reach userland.

This too happens because EF runs `SkeletoniseWrapper` through `CommandService`. I can fix both this, and the macro recording bug my having EFWrapper just call a method in `Skeletonize_`

rimadoma:
@alessandrofelder `EllipsoidOptimisationStrategyTest.testMinimumSemiAxisFilter` failed when Michael's merge commit launched the Travis build. When I ran it again manually the test passed. Can this test fail randomly?

alessandrofelder:
🤔 unfortunately don't have time to test your very good questions in detail now, but I hope to find an answer in the next 2 days.

> Would Python scripting EF still work if the images were not output parameters?

Not out of the box the way I am using it, see [Line 88 here](https://github.com/alessandrofelder/EF-helper-scripts/blob/8fa10e49e651e665550b80a7b901d6b01c607983/Python/generate-EF-for-test-images.py#L88), where I reference the output by its variable name in Java. It may be possible to access non-output variables... I don't know. What would you make it instead of an output variable?

rimadoma:
> What would you make it instead of an output variable?

@alessandrofelder Just images I explicitly display if e.g. a macro is not running. I was wondering if in Python scripts you could grab hold of open image windows like in macros / Java? Just something I'm throwing out there, but maybe the forums can help on how to control images popping open / how to do things differently depending on whether a macro is running or not.

mdoube:
> This too happens because EF runs `SkeletoniseWrapper` through `CommandService`. I can fix both this, and the macro recording bug my having EFWrapper just call a method in `Skeletonize_`

That would seem like the sensible thing to do where we have more control over the output.

PR #239, opened 2020-05-04
Add Feret point display to 3D Viewer

mdoube:
Also fixes an NPE introduced by last commits on this branch

### DIFF
--- a/Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
+++ b/Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
@@ -24,6 +24,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 package org.bonej.plugins;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
@@ -41,7 +42,6 @@ import ij.IJ;
 import ij.ImagePlus;
 import ij.ImageStack;
 import ij.measure.Calibration;
-import ij.plugin.filter.PlugInFilter;
 import marchingcubes.MCTriangulator;
 import sc.fiji.analyzeSkeleton.AnalyzeSkeleton_;
 import sc.fiji.analyzeSkeleton.SkeletonResult;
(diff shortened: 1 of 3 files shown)


Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
```diff
@@ -655,9 +655,11 @@ static ImagePlus getBinaryParticle(final int p, final ImagePlus imp, final int[]
 					point.y += yOffset;
 					point.z += zOffset;
 				}
-				surfacePoints.add(points);
 				if (points.isEmpty()) {
 					IJ.log("Particle " + p + " resulted in 0 surface points");
+					surfacePoints.add(null);
```
rimadoma, line 660: If we don't actually need to differentiate between `null` and `empty` lists (see below), then I'm not a fan of having `null` lists in the List<List<Point3f>>

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
```diff
@@ -717,42 +719,62 @@ private static double getSurfaceArea(final List<Point3f> points) {
...
+						Arrays.fill(ferets[i], Double.NaN);
 						continue;
 					}
 					
+					final int nPoints = surface.size();
+					
+					//4 triangles * 3 points for the minimal tetrahedron
+					if (nPoints < 12) {
+						Arrays.fill(ferets[i], Double.NaN);
+						throw new IllegalArgumentException("Mesh for particle "+i+" has too few points: "+nPoints);
```
rimadoma, line 750: Since you're filling the array with NaN values, it seems that a surface that's not a tetrahedron is not a show stopper. In that case let's not throw an exception, but just
```
// Since behaviour for null and empty lists is the same, let's not have 'null' lists in `surfacePoints`
if (surface == null || surfacesurface.size() < 12) {
  Arrays.fill(ferets[i], Double.NaN);
  // Can still log a strange surface here
}
```

If finding a surface with less than 12 points is truly a bug that should stop execution and throw an exception then you should see if there is one at the beginning of the method:
```
// filter(Objects::nonNull) not necessary if there are no nulls in surfacePoints
final boolean strangeSurfaces = surfacePoints.stream().filter(Objects::nonNull).anyMatch(surface -> surface.size() < 12);
if (strangeSurfaces) {
    throw new IllegalArgumentException("Bad surface!");
}
```
As the code is now only the threads that throw exceptions stop, while the rest chug happily along

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
```diff
@@ -717,42 +719,62 @@ private static double getSurfaceArea(final List<Point3f> points) {
...
+					}
+					
 					// check all the point pairs in this surface
-					ListIterator<Point3f> ita = surface.listIterator();
-					ListIterator<Point3f> itb;
-					Point3f a;
-					Point3f b;
+					Point3f a = new Point3f();
+					Point3f b = new Point3f();
```
rimadoma, line 755: Redundant initializations, can be local variables within the loops

mdoube: Agree this is neater but I worry about object creation inside hot loops. No time right now to benchmark it. Hoping that the JRE will optimise it...

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
mdoube: >  I'm not a fan of having `null` lists in the List<List>

Agree. Will avoid in the future. Better to have empty `Lists` and to detect them with `isEmpty()`.

Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
rimadoma: There's no object creation in `Point a = surface.get(p)`. A new reference variable `a` is declared, and then a reference to an object in `surface` is assigned to it. It's `Point a = new Point()` where object creation happens.

PR #244, opened 2020-05-10
Fix table list display

rimadoma:
Fixes a bug where `FractalDimensionWrapper` would display only one point table instead of one for each 3D subspace in a hyper stack.


mdoube:
Still getting this testing error:
```
[ERROR] Failures: 
[ERROR]   FractalDimensionWrapperTest.testPointResults:116 Incorrect log(count) value expected:<2.0794415416798357> but was:<1.791759469228055>
[ERROR]   FractalDimensionWrapperTest.testResults:151->lambda$testResults$4:151 Fractal dimension column has a wrong value expected:<1.4999999999999998> but was:<1.2924812503605778>
```
Maybe looking in the wrong cell of the table?

rimadoma:
I had a good reason not to do it that way. IIRC sorting columns didn't quite work, i.e. results didn't "follow" their channels/frames. However, that is no longer the case.

This fix will be bigger, and should be in another PR. I'm assuming for consistency's sake you want all the plugins to write their results in the way described above?

mdoube:
> for consistency's sake you want all the plugins to write their results in the way described above?

Yes, that would be good for users.

mdoube:
> @mdoube If you wouldn't mind adding an issue about it in the meanwhile?

OK done, see #251

PR #245, opened 2020-05-10
Move repeated code to an abstract base class

rimadoma:
* Creates an abstract base class `BoneJCommand` that all wrapper plugins extend
* Moves shared functionality to the base class

As a side note I tried moving the common output 
```
@Parameter(type = ItemIO.OUTPUT, label = "BoneJ results")
private Table<DefaultColumn<Double>, Double> resultsTable;
```
but that doesn't seem to be possible (tests started to throw `NullPointerException`).

I checked that `mvn test -PallTests` passes, and manually ran all the modern plugins.


mdoube:
Classy move, I like it.

mdoube:
> I checked that `mvn test -PallTests` passes,

I still get this failure:
```
[ERROR] Failures: 
[ERROR]   FractalDimensionWrapperTest.testPointResults:109 Incorrect log(count) value expected:<2.0794415416798357> but was:<1.791759469228055>
[ERROR]   FractalDimensionWrapperTest.testResults:144->lambda$testResults$4:144 Fractal dimension column has a wrong value expected:<1.4999999999999998> but was:<1.2924812503605778>
```
But it's unrelated to the current PR.

PR #247, opened 2020-05-11
Spring cleaning

rimadoma:
Cleans several small things. Mostly `UsageReporter`. The changes are unrelated to each other, but I'm too lazy to create several PRs.


Modern/utilities/src/main/java/org/bonej/utilities/RoiManagerUtil.java
```diff
@@ -104,15 +103,11 @@ public static boolean isActiveOnAllSlices(final int sliceNumber) {
...
 	// region -- Helper methods --
-
-	private static int clamp(final int value, final int min, final int max) {
-		if (Integer.compare(value, min) < 0) {
-			return min;
-		}
-		if (Integer.compare(value, max) > 0) {
-			return max;
+	private static int clamp(final int value, final int max) {
+		if (value < 0) {
```
mdoube: Why get rid of `min`?
Should be 
```
if ( value < min ){
    return min;
}
```

Modern/utilities/src/main/java/org/bonej/utilities/RoiManagerUtil.java
```diff
@@ -104,15 +103,11 @@ public static boolean isActiveOnAllSlices(final int sliceNumber) {
...
+			return 0;
 		}
-		return value;
+		return Math.min(value, max);
```
mdoube: I find it easier to read if all the return cases are spelled out:

```
if (value > max)
{
    return max;
}
return value;
```

Modern/utilities/src/main/java/org/bonej/utilities/RoiManagerUtil.java
mdoube: ... I realise the rationale for the change is that only 0 is ever used as `min` for this class, but I prefer to leave things explicit rather than implied.

Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/BoneJCommand.java
```diff
@@ -31,14 +32,19 @@
...
         }
         reporter.reportEvent(getClass().getName());
     }
 
+    private void initReporter() {
+        final PrefService prefService = context().getService(PrefService.class);
+        final PluginService pluginService = context().getService(PluginService.class);
+        final CommandService commandService = context().getService(CommandService.class);
+        final LogService logService = context().getService(LogService.class);
+        reporter = UsageReporter.getInstance(prefService, pluginService, commandService, logService);
```
mdoube, line 45: What happens to the logs collected by `logService`? Are they stored anywhere, and if so, saved logs need to be cleared if a user opts-out of data collection.

rimadoma:
> Much better to put output into the debug log rather than `stdout`. How can users see the log if they are curious about the data collection?

Edit > Options > ImageJ2... > SciJava log level

mdoube:
> Edit > Options > ImageJ2... > SciJava log level

OK, have to remember to update the [user docs](https://imagej.net/BoneJ2#How_is_it_implemented.3F) with this.

Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/BoneJCommand.java
rimadoma: They are printed to a window in ImageJ. The same one that shows exceptions etc. I don't think they're stored, but I don't know for sure. I'll find out.

As a side note `UsageReporter` takes in the interface `Logger`, so if we're not happy what `LogService` does we can change the actual type (or hack our own thing).

Modern/utilities/src/main/java/org/bonej/utilities/RoiManagerUtil.java
rimadoma: Agreed, the method is clearer from the outside with both parameters, I'll change it back.

mdoube: > I'll change it back.

Thanks - but change it to the ultra-simple version (i.e. not `Integer.compare()`, which obscures the purpose because to understand what is going on you have to remember which way around it works)

rimadoma: Nevermind, just noticed that the whole method was removed later in this PR as unused.

Modern/utilities/src/main/java/org/bonej/utilities/RoiManagerUtil.java
rimadoma: See above

Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/BoneJCommand.java
rimadoma: Forum drew blank so I did some digging: seems like [`StderrLogService`](https://github.com/scijava/scijava-common/blob/master/src/main/java/org/scijava/log/StderrLogService.java) is the default implementation of `LogService`. It does not store logs.

Issue #248, opened 2020-05-13
Anisotropy does not recognize image stack as 3D image


SebastienCallens:
Hello, I have an issue with some of the tools in BoneJ2, including the anisotropy tool. Whenever I try to run it on a stack of binarized images (see attached file), I get the error message that a "3D image" is required. This also happens when I try the anisotropy tool on the EF test images. 

Do you know how I should convert the image stack into a suitable 3D image for the anistropy calculation? I don't understand why it is not accepted in its current format. 

Thanks a lot in advance!

[testAnisotropy.zip](https://github.com/bonej-org/BoneJ2/files/4622338/testAnisotropy.zip)

alessandrofelder:
Hello,
Thanks for getting in touch!
The issue with the test image you have provided is that the third dimension is temporal, not spatial.
You can change this in ImageJ via `Image>Properties`.
Can I ask you to post future questions on the [ImageJ forum](https://forum.image.sc/) with a BoneJ tag, so that they will be easier to find in the future, please?

Hope this helps!

@mdoube @rimadoma we might want to specify the fact that we expect 3 _spatial_ dimensions in error message? Just a thought...

rimadoma:
> we expect 3 _spatial_ dimensions in the error message?

Not a bad idea. Do you have time to look at the EF test images?

alessandrofelder:
I am not sure which EF test images @SebastienCallens refers too?
Just tested that Anisotropy works on `shrew.tif` now, which is an EF test image givee on [bonej.org](http://bonej.org/ef), and it does :)

rimadoma:
> I am not sure which EF test images @SebastienCallens refers too?

I think images from _Plugins>BoneJ>Test images>Create EF test images_

alessandrofelder:
Ooo yes, that is a 🐛 ! Thanks for spotting that, @SebastienCallens and @rimadoma !

Slightly different to @SebastienCallens problem because that command creates extra channel dimensions (which used to be OK for EF), not extra temporal dimensions, instead of z spatial dimensions. Will fix now.

SebastienCallens:
Thanks a lot for getting back so quickly and for clarifying this issue! Regarding the EF images, I was indeed referring to the ones that @rimadoma mentioned. Just so you know, I could run my file without problems in BoneJ1. 

I will also make sure to post future questions in the forum. Thanks again!

PR #250, opened 2020-05-13
specify spatial axes for EF test images

alessandrofelder:
partially solves issue #248. EF test image creation was out of date, and created channels instead of test slices, which is fixed by this PR.

TBH, I never meant for the test image creation to reach users, so apologies for that mistake! That being said, it might be useful for others as a way of creating artificial trabecular bone-like images? We should discuss whether we want to keep it at all, but this fixes the immediate issue.


rimadoma:
Looks good. Noticed `rng.nextInt(imageDimension-1)` in many places though. The upper bound in `Random.nextInt(int)` is exclusive. Does that matter?

mdoube:
>  I never meant for the test image creation to reach users,

It should not at this stage. But, it may be helpful to include some in a more coordinated fashion, and definitely not in the main BoneJ menu: perhaps under sample images or similar.

alessandrofelder:
Agreed - taken it out for now. Not sure if it fits with the test classes either - I was using it as a command called from a script to create test images and then run EF on them, so there is no code anywhere within the BoneJ codebase that uses it. I would suggest we come back to this code (or recode it, not very complex) if and when we want to introduce a test image creation.

PS thanks for pointing the RNG bounds issue out, @rimadoma - I wasn't aware, so good to know! In short, in this case, it will just mean that the last index of the image in all three dimensions will not be considered when randomly choosing the center of an ellipsoid to draw, so won't matter that much.

PR #253, opened 2020-05-22
Update README.md

mdoube:
Point to bonej.org/legacy


rimadoma:
Modern only having wrapper plugins is not strictly accurate. E.g. most of Anisotropy down to the nitty gritty is still in BoneJ. I think the related op is too specific for imagej-ops, and still likely to change.

mdoube:
Ok, please edit to make it correct. Needs to be brief, descriptive enough, but also to remain true while the project evolves so that we don't have to come back here all the time to keep it up to date.

PR #255, opened 2020-06-01
Clean up URLs

mdoube:
- Change HTTP to HTTPS where possible
- remove broken links
- update dx.doi.org to doi.org

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,13 +3,13 @@
 ## Prerequisites
 * Familiarize yourself with GitHub
 * [Java 8]() (newer version is fine, but the project still uses 1.8 bytecode)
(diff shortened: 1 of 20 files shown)


mdoube:
There are quite a few `http://` URLs in `pom.xml`:
```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
I did not touch this boilerplate stuff, because I wasn't sure what the effect of changing to `https://` would be and it probably ought to be applied wholesale across the ImageJ/Fiji project if it is a Good Idea.

PR #257, opened 2020-06-05
Calibrate detailed skeleton branch output.

### LINKED ISSUE #256: Some AnalyzeSkeleton detailed branch results are not calibrated
**Describe the bug**
Detailed branch information reported by BoneJ's AnalyzeSkeleton and Skeleton option in Particle Analyser is uncalibrated

**To Reproduce**
1. Run AnalyzeSkeleton on a skeletonised image that has spatial calibration, selecting detailed branch information
2. Run Analyse Skeleton from BoneJ's menus on the same image
3. Note that the Euclidean Distance and branch location values are different

**Expected behavior**
Output from BoneJ's calls to AnalyzeSkeleton's methods should exactly match output from AnalyzeSkeleton

**Additional context**
Reported here:
https://forum.image.sc/t/skeleton-branch-length-producing-zero-values-for-most-of-the-soil-pores-in-bonej2/38505/6

mdoube:
Fixes #256

Updates the test to use a calibrated input image.

### DIFF
--- a/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/AnalyseSkeletonWrapper.java
+++ b/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/AnalyseSkeletonWrapper.java
@@ -34,6 +34,7 @@ import static org.bonej.wrapperPlugins.wrapperUtils.Common.cancelMacroSafe;
 import ij.ImagePlus;
 import ij.ImageStack;
 import ij.gui.Roi;
+import ij.measure.Calibration;
 
 import io.scif.FormatException;
 import io.scif.services.FormatService;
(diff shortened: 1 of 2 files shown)


Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/AnalyseSkeletonWrapper.java
```diff
@@ -319,15 +321,18 @@ private void showAdditionalResults(final SkeletonResult results) {
...
-				final double distance = MathArrays.distance(new double[] { point.x,
-					point.y, point.z }, new double[] { point2.x, point2.y, point2.z });
+				((DoubleColumn) columns.get(6)).add((point2.x) * cal.pixelWidth);
+				((DoubleColumn) columns.get(7)).add((point2.y) * cal.pixelHeight);
+				((DoubleColumn) columns.get(8)).add((point2.z) * cal.pixelDepth);
+				final double distance = MathArrays.distance(
+					new double[] { point.x * cal.pixelWidth,
+						point.y * cal.pixelHeight, point.z * cal.pixelDepth },
+					new double[] { point2.x * cal.pixelWidth, 
+						point2.y * cal.pixelHeight, point2.z * cal.pixelDepth });
```
rimadoma, line 335: Could have a helper method
```
// Calibration is a private field
	private Vector3dc getCalibratedPoint(Vertex vertex) {
		final Point point = vertex.getPoints().get(0);
		final double px = point.x * calibration.pixelWidth;
		final double py = point.y * calibration.pixelHeight;
		final double pz = point.z * calibration.pixelDepth;
		return new Vector3d(px, py, pz);
	}
```
Then as a bonus you'd get distance just from `point1.distance(point2)`

mdoube: > Could have a helper method

`AnalyzeSkeleton3D_` has exactly what you suggest, but it's private (and inefficiently implemented with `Math.pow`). I made a PR to fix that, https://github.com/fiji/AnalyzeSkeleton/pull/16

I can see that this method should be both `static` and maybe in another class entirely, but refactoring would open a whole can of worms that none of us can prioritise right now.

PR #262, opened 2020-06-14
Explicity create and protect the background particle identity

### LINKED ISSUE #261: Particle Analyser throws lots of errors when there is only one particle at the end of the stack
**Describe the bug**
This is a super weird bug. Seems to be out-by-one and 'null' type of errors when there are only sparse or one particle, at the end of the stack.

See:
https://forum.image.sc/t/null-pointer-exception-in-particle-analyser/38946

Root cause of some of it seems to be that `ConnectedComponents.getNParticles()` sometimes returns one too few particles, which borks all the arrays that rely on there being an index in the arrays for every ID in the label stack.

mdoube:
This PR fixes #261, which resulted from an implicit creation of the background particle ID (0) that was either not created, or created and then destroyed when there was no foreground particle in the zeroth image processing chunk.


mdoube:
Merge this PR before #263

PR #263, opened 2020-06-15
Retain edge-touching background particles when purifying

### LINKED ISSUE #259: Purify removes background particles that are connected to the outside (regression)
**Describe the bug**
When a stack is purified, foreground particles that are touching the edges but not the largest particle are removed, but background particles touching the edges but not the biggest background particle should be retained. In other words, background particles touching edges should be attached to each other (the 'outside' should be considered one large piece of background).

This was handled properly in BoneJ1 but the new particle analyser seems to have reintroduced the wrong behaviour.

It can probably be fixed by making sure the background edge particles are assigned to the same particle. Note that `touchEdges()` is missing from new `Purify` but present in old `Purify`, which is the likely source of this bug.

**To Reproduce**
1. Download umzc_378p_Apteryx_haastii_head.tif.bz2 from https://doi.org/10.6084/m9.figshare.7257179
2. Extract umzc_378p_Apteryx_haastii_head.tif from the .bz2 file
 - note that it is a 16-bit greyscale X-ray microtomography image with isotropic pixel spacing (10.1 µm), and that the trabecular bone and marrow fills the entire image volume
3. Run Process > Filters > Gaussian Blur (3D) with a sigma of 2.0 in all 3 dimensions.
4. Run Image > Adjust > Threshold, select Dark background and Stack histogram, click Auto and then click Apply
5. In the Convert Stack to Binary window, unselect all options, Method: Default, Background: Dark
7. Run Plugins > BoneJ > Purify to remove small background particles

**Expected behavior**
Background particles touching image edges should be retained in the purified image.

**Screenshots**
Correct behaviour is on the left (from BoneJ 1.4.3) and wrong behaviour is on the right (from BoneJ 7.0.6)
Screenshot from 2020-06-11 13-44-51

mdoube:
Fixes #259

### DIFF
--- a/Legacy/bonej/src/main/java/org/bonej/plugins/ConnectedComponents.java
+++ b/Legacy/bonej/src/main/java/org/bonej/plugins/ConnectedComponents.java
@@ -168,7 +168,7 @@ public class ConnectedComponents {
 	private static int[][] generateLut(ArrayList<MutableList<IntHashSet>> chunkMaps, int[] chunkIDOffsets) {
 		// merge labels between the HashSets, handling the chunk offsets and indexes
 		bucketFountain(chunkMaps, chunkIDOffsets);
-
+		
 		HashMap<Integer, Integer> lutMap = makeLutMap(chunkMaps);
 
 		return lutFromLutMap(lutMap, chunkMaps, chunkIDOffsets);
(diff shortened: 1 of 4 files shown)


mdoube:
Please merge PR #262 to master first.

PR #264, opened 2020-06-16
Add LICENSE.txt to each module

mdoube:
The release script automatically generated `LICENSE.txt` files, one for each module `JAR`.


mdoube:
> Just so I understand: which release script?

This one: https://github.com/scijava/scijava-scripts/blob/master/release-version.sh

PR #272, opened 2020-10-27
Improvements to EF user experience, bug fixes to Particle Analyser & Slice Geometry

mdoube:
This PR addresses issues with EF being a bit too verbose for most users, fixes a rarely-encountered bug in Particle Analyser, and restores Imax and Imin nomenclature to the previous in Slice Geometry (min & max refer now to the principal axes around which the moments are calculated, which has the effect that Imin > Imax in most cases).


alessandrofelder (approved):
Looks good to me!

PR #274, opened 2020-10-30
Dependency updates

mdoube:
Update pom.xml to use scijava 29.2.1 and handle downstream effects of that change.

### DIFF
--- a/Legacy/bonej/pom.xml
+++ b/Legacy/bonej/pom.xml
@@ -123,7 +123,6 @@
         <dependency>
             <groupId>sc.fiji</groupId>
             <artifactId>LocalThickness_</artifactId>
-            <version>${LocalThickness_.version}</version>
         </dependency>
         <dependency>
             <groupId>sc.fiji</groupId>
--- a/Legacy/bonej/pom.xml
+++ b/Legacy/bonej/pom.xml
@@ -148,12 +148,12 @@
         <dependency>
             <groupId>org.eclipse.collections</groupId>
             <artifactId>eclipse-collections-api</artifactId>
-            <version>10.2.0</version>
+            <version>10.4.0</version>
         </dependency>
         <dependency>
             <groupId>org.eclipse.collections</groupId>
             <artifactId>eclipse-collections</artifactId>
-            <version>10.2.0</version>
+            <version>10.4.0</version>
         </dependency>
 
         <!-- Test dependencies -->
--- a/Legacy/bonej/pom.xml
+++ b/Legacy/bonej/pom.xml
@@ -147,12 +147,10 @@
         <dependency>
             <groupId>org.eclipse.collections</groupId>
             <artifactId>eclipse-collections-api</artifactId>
-            <version>10.4.0</version>
         </dependency>
         <dependency>
             <groupId>org.eclipse.collections</groupId>
             <artifactId>eclipse-collections</artifactId>
-            <version>10.4.0</version>
         </dependency>
 
         <!-- Test dependencies -->
--- a/Modern/ops/pom.xml
+++ b/Modern/ops/pom.xml
@@ -112,13 +112,6 @@
         <dependency>
             <groupId>net.imagej</groupId>
             <artifactId>imagej-ops</artifactId>
-            <!-- Try removing after pom-scijava 29.x.x -->
-            <exclusions>
-                <exclusion>
-                    <groupId>com.miglayout</groupId>
-                    <artifactId>miglayout</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
         <dependency>
             <groupId>net.imagej</groupId>
@@ -148,13 +141,6 @@
             <groupId>org.joml</groupId>
             <artifactId>joml</artifactId>
         </dependency>
-        <!-- Needed for UI. Try removing after pom-scijava 29.x.x -->
-        <dependency>
-            <groupId>com.miglayout</groupId>
-            <artifactId>miglayout-swing</artifactId>
-            <version>5.2</version>
-            <scope>runtime</scope>
-        </dependency>
 
         <!-- Testing dependencies -->
         <dependency>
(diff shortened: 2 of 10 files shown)


alessandrofelder (commented):
I am worried about commenting out tests.
I assume the reasoning behind this is that they have stopped passing?
I will try and figure out what the problem is this week.

mdoube:
> I am worried about commenting out tests.

Me too. These tests seem like obscure corner cases rather than main function (e.g. if file is not an image), so I'm not too worried.

> I will try and figure out what the problem is this week.

OK but balance the cost of your time against the risk that these tests are meant to be averting. The Fractal Dimension one is a known unknown - a kind of Heisentest. Sometimes passes, sometimes doesn't. The Analyse Skeleton one seems obscure, like a file format test which should be handled way upstream from the wrapper plugin, and could possibly be discarded for this reason. It's not obvious to me what it's actually testing, and whether what it is testing is relevant to this level of plugin.

alessandrofelder:
Starting on the `FindRidgePoints` test - I am getting the tests to pass but `mvn` showing 
```
[ERROR] Cannot create plugin: class='io.scif.convert.FileToDatasetConverter', priority=1.0, enabled=true, pluginType=Converter
```
Is this what you are seeing too, @mdoube ?

mdoube:
> Is this what you are seeing too, @mdoube ?

Yes exactly. Test passes when run directly in Eclipse, but throws this to the console:
```
[ERROR] Cannot create plugin: class='io.scif.convert.FileToDatasetConverter', priority=1.0, enabled=true, pluginType=Converter
```

mdoube:
`ElementUtilTest.testIsBinaryReturnsFalseIfIntervalEmpty()` fails with:

```
net.imglib2.exception.InvalidDimensionsException: Expected only positive dimensions but got: [0]
	at net.imglib2.Dimensions.verifyAllPositive(Dimensions.java:166)
	at net.imglib2.Dimensions.verify(Dimensions.java:208)
	at net.imglib2.img.array.ArrayImgFactory.create(ArrayImgFactory.java:90)
	at net.imglib2.img.array.ArrayImgFactory.create(ArrayImgFactory.java:69)
	at net.imglib2.img.array.ArrayImgs.doubles(ArrayImgs.java:555)
	at org.bonej.utilities.ElementUtilTest.testIsBinaryReturnsFalseIfIntervalEmpty(ElementUtilTest.java:234)
```
Which suggests to me that this condition (an image with zero size) is now checked for and prevented earlier in execution.
Propose deleting this test.

mdoube:
`AnalyseSkeletonWrapperTest.testBadFormatIntensityImageCancelsPlugin()` fails with:

```
org.junit.ComparisonFailure: Cancel reason is incorrect expected:<[Image format is not recognized]> but was:<[An error occurred while opening the image]>
	at org.junit.Assert.assertEquals(Assert.java:125)
	at org.bonej.wrapperPlugins.AnalyseSkeletonWrapperTest.testBadFormatIntensityImageCancelsPlugin(AnalyseSkeletonWrapperTest.java:170)
```
It's not clear to me what error condition the test is actually checking for:

```
	@Test
	public void testBadFormatIntensityImageCancelsPlugin() throws Exception {
		// SETUP
		final ImagePlus imagePlus = IJ.createImage("test", 3, 3, 3, 8);
		final File exceptionFile = mock(File.class);
		when(exceptionFile.getAbsolutePath()).thenReturn("file.foo");
		when(MOCK_UI.chooseFile(any(File.class), anyString())).thenReturn(
			exceptionFile);

		// EXECUTE
		final CommandModule module = command().run(
			AnalyseSkeletonWrapper.class, true, "inputImage", imagePlus,
			"pruneCycleMethod", "Lowest intensity voxel").get();

		// VERIFY
		assertTrue("Plugin should have cancelled", module.isCanceled());
		assertEquals("Cancel reason is incorrect", "Image format is not recognized",
			module.getCancelReason());
	}
```

alessandrofelder:
I think that `AnalyseSkeletonWrapperTest.testBadFormatIntensityImageCancelsPlugin()` checks that the `AnalyseSkeletonWrapper` throws a `FormatException` when the input file name extension is `.foo`. This is to make sure that the user gets a different, more nuanced error message when they pass an image with a dodgy file name extension than in cases where something else goes wrong when opening the image. So for that reason, I think having the test would be useful.

The underlying cause of this not passing anymore is the `DefaultFormatService`'s `getFormat()` method now returns `null` instead of throwing a `FormatException`, as far as I can tell.

I hope to dig deeper tomorrow.

mdoube:
> The underlying cause of this not passing anymore is the `DefaultFormatService`'s `getFormat()` method now returns `null` instead of throwing a `FormatException`, as far as I can tell.

Right, so the `null` then triggers somewhere a `NullPointerException`:

In `AnalyseSkeletonWrapper.openIntensityImage()`
```
		catch (final IOException | NullPointerException e) {
			cancelMacroSafe(this, "An error occurred while opening the image: "+e.toString());
			logService.trace(e);
		}
```
Is logged like this by the test:
```
An error occurred while opening the image: java.io.FileNotFoundException: file.foo (No such file or directory)
```

alessandrofelder:
Exactly.
The NPE is thrown because the `File.foo` doesn't actually exist (it's just being mocked by the test to check that the FormatException is thrown). So the code thinks the format is fine (no format exception) and then tries to open the file?

[This set of changes](https://github.com/scifio/scifio/commit/f8d6868f59a9f8c1af49588201063186f8fb9e8a) to the DefaultFormatService is the key. 

This is probably my lack of understanding what an empty `FormatList` means here, but possibly worth double-checking: @ctrueden @frauzufall, are you happy with returning `null` when the FormatList is empty, or should that trigger a FormatException too?
(I am basing this on the fact that in the [FormatService javadoc](https://github.com/scifio/scifio/blob/bdb8ab764d58243d9f3a9f3936d468d695268a9c/src/main/java/io/scif/services/FormatService.java#L186), it says that `getFormat` will throw a FormatException
"if no compatible format was found, or if something goes wrong checking the source".
)
Thanks a lot for your help!

alessandrofelder:
I suggest we remove the negative test from AnalyzeSkeletonWrapper for now, and I've opened an issue so we don't forget to reinstate it once we figure out how to do it. See #275. At least this way, we don't get held up.

alessandrofelder (approved):

Issue #276, opened 2020-11-17
First pass particle labelling runs out of label space on huge images on big machines


mdoube:
**Describe the bug**
Particle Analyser fails to complete properly, complaining for the last thread:

```
Exception in thread "Thread-132" java.lang.ArrayIndexOutOfBoundsException: 2016
	at org.bonej.plugins.ConnectedComponents.lambda$firstIDAttribution$1(ConnectedComponents.java:264)
	at java.lang.Thread.run(Thread.java:748)
```
and for many threads:

```
Exception in thread "Thread-126" java.lang.IllegalArgumentException: ID 7602176 is greater than the allowed range (max 7602175)
	at org.bonej.plugins.ConnectedComponents.expandMap(ConnectedComponents.java:998)
	at org.bonej.plugins.ConnectedComponents.lambda$firstIDAttribution$1(ConnectedComponents.java:323)
	at java.lang.Thread.run(Thread.java:748)
```

**To Reproduce**
1. Open a huge image, e.g. 150 GB with 3 million particle on a big machine e.g. 132 CPU threads
2. Run Particle Analyser

**Expected behavior**
Particle Analyser completes successfully

**Environment**
 - OS: Windows 10
 - Java version ??
 - 64 Intel Xeon Gold 6142M processors @2.60 GHz base speed. Nvidia Quadro RTX 8000 GPU, and 2 TB RAM

**Additional context**
First reported here:
https://forum.image.sc/t/bonej-analyse-particles-file-size-limitation/45137/7

PR #278, opened 2020-11-17
Expand the allowable label space to use all positive int values

mdoube:
Addresses issue #276

Particle labels were previously capped at 2<sup>23</sup> which is the limit of integer precision represented by `float`. This limit was exceeded by large images (e.g. 150 GB) with many particles (3 million) running on computers with many threads (e.g. 132).

This solution allows the entire range of positive `int` (up to 2<sup>32</sup>) to be used during particle labelling and analysis of the resulting label array. The limitation of **displaying** labels in ImageJ1 as 32-bit `float` images remains.


alessandrofelder (approved):
The fix makes sense to me. Is it worth documenting the display behaviour in the javadoc as well as in the log? (I think probably yes?)

PR #280, opened 2021-01-06
Quit Slice Geometry if a non-rectangular ROI is provided

### LINKED ISSUE #279: Slice Geometry throws an AIOOB Exception if the ROI falls outside image bounds
**Describe the bug**
Slice Geometry throws an AIOOB Exception if the ROI falls outside image bounds

```
java.lang.ArrayIndexOutOfBoundsException: -14969
	at ij.process.ByteProcessor.get(ByteProcessor.java:251)
	at org.bonej.plugins.SliceGeometry.calculateCentroids(SliceGeometry.java:506)
	at org.bonej.plugins.SliceGeometry.run(SliceGeometry.java:323)
	at ij.IJ.runUserPlugIn(IJ.java:235)
	at ij.IJ.runPlugIn(IJ.java:198)
	at ij.Executer.runCommand(Executer.java:150)
	at ij.Executer.run(Executer.java:68)
	at java.lang.Thread.run(Thread.java:748)
```

**To Reproduce**
1. Open a greyscale image
2. Set an ellipse ROI
3. Shift the ROI so that some of it falls outside the image boundary
4. Run _Slice Geometry_

**Expected behavior**
Slice Geometry runs without an Exception

**Additional context**
It is impossible to specify a rectangular ROI that falls outside the image boundaries.
[Reported here.](https://forum.image.sc/t/second-moment-of-area-imin-and-imax/47098/8)

mdoube:
Fixes #279

### DIFF
--- a/Legacy/bonej/src/main/java/org/bonej/plugins/SliceGeometry.java
+++ b/Legacy/bonej/src/main/java/org/bonej/plugins/SliceGeometry.java
@@ -225,6 +225,12 @@ public class SliceGeometry implements PlugIn, DialogListener {
 			IJ.noImage();
 			return;
 		}
+		Roi roi = imp.getRoi();
+		if (roi == null) {}
+		else if (roi.getType() != Roi.RECTANGLE) {
+			IJ.showMessage("ROI Error", "Slice Geometry expects only one rectangular ROI");
+			return;
+		}
 		cal = imp.getCalibration();
 		vW = cal.pixelWidth;
 		vH = cal.pixelHeight;


alessandrofelder (approved):
Seems reasonable to me.

Legacy/bonej/src/main/java/org/bonej/plugins/SliceGeometry.java
```diff
@@ -225,6 +225,12 @@ public void run(final String arg) {
 			IJ.noImage();
 			return;
 		}
+		Roi roi = imp.getRoi();
+		if (roi == null) {}
```
rgaiacs, line 229: Can we add

```
IJ.showMessage("ROI Error", "Slice Geometry expects one rectangular ROI but none was found");
return;
```

inside the `if`-clause block?

PR #281, opened 2021-02-03
Reduce pixel and array lookups to get a c. 35% reduction in execution time

mdoube:
This is a small performance enhancement that reduces by half the number of times pixels have to be looked up from the image stack. Also some dead code is removed. Same was suggested [here](https://github.com/maarzt/imagej-ops/commit/5e1ef886b3f84ea278c5710c0d89638f9fcaaba8#diff-a06b9f93651e2b6be331719c7f00d7af4a3460fc8dff3987dcc5edfa9b20c9fd) for Modern Connectivity, but so far I haven't been able to get that to compile & pass tests.
Array lookups for the octant table are removed in favour of 8 primitive `byte`.

### DIFF
--- a/Legacy/bonej/src/main/java/org/bonej/plugins/Connectivity.java
+++ b/Legacy/bonej/src/main/java/org/bonej/plugins/Connectivity.java
@@ -227,10 +227,11 @@ public class Connectivity implements PlugIn {
 			threads[thread] = new Thread(new Runnable() {
 				@Override
 				public void run() {
+					final byte[] octant = new byte[9];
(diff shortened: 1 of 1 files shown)


alessandrofelder (approved):
If I understand correctly, the reduction in execution time (yay!) stems from re-using entries in the `octant` byte array from a previous call to `getOctant`? So that means `getOctant` is only useable in a specific order, which ensures that the entries that it's re-using are correct? If so, I would suggest to document this, so we remember not to call `getOctant` wrongly in the future? (I get that it's unlikely with it being a private function and all, but maybe still good practice?)

Assuming I've got the above right, I am happy with the change.

PR #283, opened 2021-02-05
Update scijava to 30.0.0

mdoube:

PR #285, opened 2021-02-10
Usage debug

mdoube:
Usage data were not being collected for Legacy plugins when debug mode was off.

### DIFF
--- a/Legacy/bonej/src/main/java/org/bonej/plugins/UsageReporter.java
+++ b/Legacy/bonej/src/main/java/org/bonej/plugins/UsageReporter.java
@@ -157,25 +157,21 @@ public final class UsageReporter {
 				utmhid + utmr + utmp + utmac + utmcc);
 			final URLConnection uc = url.openConnection();
 			uc.setRequestProperty("User-Agent", userAgentString());
-			if (IJ.debugMode) {
-				logRequest(url, uc);
+			try (final BufferedReader reader = new BufferedReader(new InputStreamReader(
+				uc.getInputStream()))) {
+					if (IJ.debugMode) {
+						IJ.log(url.toString());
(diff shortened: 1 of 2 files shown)


mdoube:
This PR should be approved and merged after #283 due to the scijava version bump commit that is hanging around here.

Legacy/bonej/src/main/java/org/bonej/plugins/UsageReporter.java
```diff
@@ -157,8 +157,15 @@ public void send() {
...
+			try (final BufferedReader reader = new BufferedReader(new InputStreamReader(
+				uc.getInputStream()))) {
+					if (IJ.debugMode) {
+						IJ.log(url.toString());
+						IJ.log(uc.getRequestProperty("User-Agent"));
+						reader.lines().forEach(IJ::log);
+					}
+			} catch (final IOException e) {
+				IJ.error(e.getMessage());
 			}
```
alessandrofelder, line 169: I am probably being silly but I don't understand how the changes make a difference.

The only thing I see changing when `IJ.debugMode` is `false` is that the `reader` variable gets constructed, and then goes out of scope at the end of the `send()` function?

mdoube: Right - no idea either. I knocked out some lines to see what is essential and it's this:
```
final BufferedReader reader = new BufferedReader(new InputStreamReader(uc.getInputStream()))
```
Which I interpret to mean that the request isn't actually sent until the `BufferedReader` tries to read from the `URLConnection`.

alessandrofelder (approved):
Might be worth sticking a comment in there saying that the construction of the `BufferedReader` seems to be essential as that isn't clear from reading the code on its own. Otherwise happy with this.

PR #289, opened 2021-02-27
Purify gracefully handles the situation when there are no particles

mdoube:
Previously it threw an exception if no particles were found.

Adds test cases for no particles, multiple particles and a single
particle in multiple locations.


rgaiacs (approved):
`Purify.java` looks good to me. I need to be study ij.ImageStack to better understand what is the test script is doing.

mdoube:
> `Purify.java` looks good to me. I need to be study ij.ImageStack to better understand what is the test script is doing.

You could start here:
https://imagej.nih.gov/ij/developer/api/ij/ImageStack.html

Source is here:
https://github.com/imagej/imagej1/blob/master/ij/ImageStack.java

Test is adding 2D image slices containing only background (0) to a 3D image stack and then drawing solid cuboid boxes of foreground (255) in the stack, of specified size and location, then Purifying and doing a particle analysis to check that only one particle remains after Purify.

PR #296, opened 2021-03-30
Use multithreaded bucket fountain only when there are >1m particles

mdoube:
The cost involved in running many threads does not pay off for smaller images with few particles, so this version 'changes gears' from single threaded to multithreaded once a cut point (1×10⁶) of particle first labels is reached.

This change also means that first labels are counted more efficiently and earlier, which removes the need to iterate over the `IntHashSets` in `makeLutMap()`.


alessandrofelder (approved):
Looks good to me.

Issue #295, opened 2021-03-24
Move user documentation links from imagej.net/BoneJ to https://imagej.github.io/BoneJ


mdoube:
**Describe the bug**
ImageJ Wiki docs have been / are being moved from imagej.net to https://imagej.github.io/

Links need to be updated. Done so far:
- bonej.org
- README.md (#294)

But need to update links elsewhere such as the 'Help' buttons in some plugins' dialogs.

**Additional context**
https://forum.image.sc/t/imagej-wiki-is-read-only/39672

rgaiacs:
`grep -ir 'imagej.net' *` provides

- [ ] `CONTRIBUTING.md:* [Maven](https://imagej.net/Maven)`
- [ ] `CONTRIBUTING.md:* [Git](https://imagej.net/Git)`
- [ ] `CONTRIBUTING.md:1) Create a [fork](https://imagej.net/How_to_contribute_to_an_existing_plugin_or_library)`
- [ ] `Legacy/util/pom.xml:    <url>https://imagej.net/BoneJ_experimental</url>`
- [ ] `Legacy/util/pom.xml:            <url>https://imagej.net/User:Rdom</url>`
- [ ] `Legacy/util/pom.xml:            <url>https://imagej.net/User:Rueden</url>`
- [ ] `Legacy/util/pom.xml:            <url>https://imagej.net/User:Hinerm</url>`
- [ ] `Legacy/bonej/pom.xml:    <url>https://imagej.net/BoneJ_experimental</url>`
- [ ] `Legacy/bonej/pom.xml:            <url>https://imagej.net/User:Rdom</url>`
- [ ] `Legacy/bonej/pom.xml:            <url>https://imagej.net/User:Rueden</url>`
- [ ] `Legacy/bonej/pom.xml:            <url>https://imagej.net/User:Hinerm</url>`
- [ ] `Legacy/bonej/src/main/java/org/bonej/DebugLauncher.java:        final ImagePlus image = IJ.openImage("https://imagej.net/images/bat-cochlea-volume.zip");`
- [ ] `Legacy/bonej/src/main/java/org/bonej/plugins/SliceGeometry.java:		gd.addHelp("https://imagej.net/BoneJ2#Slice_geometry");`
- [ ] `Legacy/bonej/src/main/java/org/bonej/plugins/Connectivity.java: *      Skeletonize3D_ plugin: <a href="https://imagej.net/Skeletonize3D">`
[...]

PR #299, opened 2021-04-23
Update documentation

rgaiacs:
Update links (see https://github.com/bonej-org/BoneJ2/issues/295)

`Legacy/bonej/src/main/java/org/bonej/DebugLauncher.java` was not update because `https://imagej.net/images/bat-cochlea-volume.zip` might be already in the system.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,14 +2,14 @@
 
 ## Prerequisites
 * Familiarize yourself with GitHub
-* [Java 8]() (newer version is fine, but the project still uses 1.8 bytecode)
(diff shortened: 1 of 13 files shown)


mdoube (approved):
Looks good

mdoube:
> Update links (see #295)
> 
> `Legacy/bonej/src/main/java/org/bonej/DebugLauncher.java` was not update because `https://imagej.net/images/bat-cochlea-volume.zip` might be already in the system.

I uploaded `bat-cochlea-volume.zip` to https://imagej.github.io/images/bat-cochlea-volume.zip so you could update the link there.

PR #303, opened 2022-04-06
Show a warning in the log rather than in a UI dialog.

mdoube:
Logging with LogService is intended to reduce dependence on there being a live UI, so that the wrapper can be called from a headless script more reliably.


mdoube:
Should close #304

alessandrofelder (approved):
Looks good to me!

PR #306, opened 2022-04-08
Delete release.properties

mdoube:
GitHub Actions build fails with the message
```
  Error:  Spurious release.properties file is present
  Remove the file from version control and try again.
```
This commit removes the `release.properties` file to try again.


mdoube:
Build now triggers and passes.

ctrueden (approved):
@mdoube Yep, that file should only exist on release tags, not on the mainline branch. It indicates that the revision is intended to be built and deployed as a release via `mvn release:perform`. Logic is [here](https://github.com/scijava/scijava-scripts/blob/f9d592c558214c5fe2fc75a3584a5ceb53badd81/ci-build.sh#L196-L211) if you're curious.

PR #313, opened 2022-04-12
Refactor image anisotropy check to handle headless execution.

mdoube:
The anisotropy check previously always showed a GUI if the image was anisotropic, but this broke headless execution. If code is called from a headless environment, e.g. a Python script, a warning is printed to the console and execution continues, even if results may be unreliable.

### DIFF
--- a/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/wrapperUtils/Common.java
+++ b/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/wrapperUtils/Common.java
@@ -118,7 +118,7 @@ public final class Common {
 
 	/**
 	 * Shows a warning dialog about image anisotropy, and asks if the user wants
-	 * to continue. If the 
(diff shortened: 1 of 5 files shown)


Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/wrapperUtils/Common.java
```diff
@@ -118,22 +118,29 @@ public static <C extends ComplexType<C>> ImgPlus<BitType> toBitTypeImgPlus(
 
 	/**
 	 * Shows a warning dialog about image anisotropy, and asks if the user wants
-	 * to continue.
+	 * to continue. If the
```
alessandrofelder, line 121: Comment sentence needs completing? sth like:
```suggestion
	 * to continue. If the plugin is running in headless mode, execution continues anyway, and a warning is printed to the log.
```

alessandrofelder (approved):
Small suggestion about a comment, but otherwise looks good.

Issue #304, opened 2022-04-06
Headless operation of IntertrabecularAngles is blocked by a a warning wanting a GUI


mdoube:
**Describe the bug**
Intertrabecular Angles cannot run in headless mode when there are more than one skeletons because a GUI dialog is triggered.

See: 
https://forum.image.sc/t/save-tiff-stack-from-ellipsoid-factor-in-headless-mode/58811/59

PR #314, opened 2022-04-13
Intertrabecular Angles: optionally show the skeleton image using Parameter conventions

mdoube:
Refactors the output behaviour of the ITA plugin wrapper so that whether a skeleton image is shown is under total control of the user option. Both user option and image output are specified as Parameters conforming to the IJ2 plugin design language.


alessandrofelder (approved):
LGTM 🎉

mdoube:
Addresses #304

PR #322, opened 2022-08-27
SciJava version bump to 32.0.0

mdoube:


ctrueden:
@mdoube I looked through BoneJ2's POM files, and don't see any reason not to bump to 32.0.0 straight away as done in this PR. LGTM.

ctrueden (approved):

PR #323, opened 2023-02-07
Changed method visibility in Ellipsoid

sjcross:
Hi,

This commit just changes the visibility of the getXMinAndMax method in the Ellipsoid class to public, so it matches getYMinAndMax and getZMinAndMax and can be used elsewhere.

Cheers,
Stephen


mdoube:
This fix is in 7.0.15, released to the ImageJ update site just now.
https://maven.scijava.org/#nexus-search;gav~org.bonej~~7.0.15~~

PR #325, opened 2023-03-22
Use the isotropy checker in AxisUtils for calibrating surfaces

### LINKED ISSUE #324: Surface Area always complains about unmatching axis calibration even when they are perfectly matched
**Describe the bug**
Regardless of spatial calibration isotropy, Surface Area always complains that axes are not calibrated.

**To Reproduce**
1. Run Surface Area with 'Export STL' checked.

**Expected behavior**
Surface Area accepts isotropic calibration or uncalibrated images and applies calibration to the results, without a warning.

mdoube:
The axis checker in Surface Area is redundant because the same functionality is provided in AxisUtils.

Fixes #324


alessandrofelder (approved):
LGTM 🚀

PR #329, opened 2023-05-10
Set the STL output location during setup by using @Parameter annotation.

mdoube:
This should make it possible to use Surface Area from a headless script, by passing the output directory as an argument in the CommandService.run() method. The rest of the file name is generated automatically from the input file name + a submesh number (if needed) + .stl.

Reported here:
https://forum.image.sc/t/bonej-thickness-macro-wont-work-using-pyimagej/80704/8

### DIFF
--- a/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/SurfaceAreaWrapper.java
+++ b/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/SurfaceAreaWrapper.java
@@ -30,7 +30,6 @@
 
 package org.bonej.wrapperPlugins;
 
-import static org.bonej.utilities.Streamers.spatialAxisStream;
 import static org.bonej.wrapperPlugins.CommonMessages.NOT_3D_IMAGE;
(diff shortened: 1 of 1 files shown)


alessandrofelder (approved):
Code changes make sense to me :+1: 

Have you tried running this from a script locally @mdoube as a sanity check? (if not, would that be worth trying? Your formulation of "should" in the PR description worried me :laughing: )

mdoube:
> Have you tried running this from a script locally @mdoube as a sanity check? (if not, would that be worth trying? Your formulation of "should" in the PR description worried me laughing )

Yes! But not headless

```python
# @CommandService cs
from ij import IJ
from net.imglib2.img import ImagePlusAdapter

imp = IJ.openImage("/home/mdoube/Desktop/small_osteocyte_lacunae_binary.tif")
img = ImagePlusAdapter.wrapImgPlus(imp)
outputDirectory = "/home/mdoube/Desktop/"

wrapper = cs.run("org.bonej.wrapperPlugins.SurfaceAreaWrapper", False, ["inputImage", img, "exportSTL", True, "stlDirectory", outputDirectory])
```

PR #333, opened 2023-06-15
Broaden the data type of EllipsoidFactor input images

gselzer:
Ran into this issue working on a [new napari-imagej use case](https://github.com/imagej/napari-imagej/pull/240)

`EllipsoidFactorWrapper` [expects an input image of `UnsignedIntType`](https://github.com/bonej-org/BoneJ2/blob/2683c50e8990a6deb767f3c954445fcee52e942e/Modern/wrapperPlugins/src/main/java/org/bonej/wrapperPlugins/EllipsoidFactorWrapper.java#L151), but funnily enough it can be run with images of many types that are not of this data type. I didn't see any real need for the input image data type to be `UnsignedIntType`, so I broadened the type out to any native real type.


alessandrofelder:
Looks good to me! Thanks @gselzer .

alessandrofelder (approved):

PR #338, opened 2023-07-10
Fix an int overflow caused by int multiplications

mdoube:
Maths as written are processed as ints meaning it overflows before getting to the doubles. Casting to long at the beginning fixes it.

The result was that for images with side lengths > 1300 pixels, Conn.D would be reported as negative, which is physically impossible because it implies a negative volume that cannot exist in this universe, when Connectivity itself is positive.


alessandrofelder (approved):
👍 
(Might be good to add a test for this someday)

PR #340, opened 2023-09-29
Filter out redundant surface mesh vertices

mdoube:
Triangles in the surface mesh share vertices but the vertices are repeated for every triangle in the mesh. This change reduces the number of points for the surface by 4-6× and should in principle reduce the time for brute-force all-to-all calculations used by Feret max by 16-36×. In testing on small meshes there is a several-fold speed up but not as much as predicted.

### DIFF
--- a/Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
+++ b/Legacy/bonej/src/main/java/org/bonej/plugins/ParticleAnalysis.java
@@ -34,6 +34,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


alessandrofelder (approved):
Nice. Changes LGTM.
Just as a comment because I don't know the PA code well: we're sure that no other part of the code relies on having those points more than once?

mdoube:
> Nice. Changes LGTM. Just as a comment because I don't know the PA code well: we're sure that no other part of the code relies on having those points more than once?

Surface Area calculation requires the triangles to be complete, so it doesn't call the filtering method. Only ellipsoid fitting and feret max, which don't care about triangles, use the filtering method.

PR #343, opened 2024-04-05
Remove usage reporter classes and tests

mdoube:
Remove all usage reporting classes and calls to them, and remove tests that refer to them.

The BoneJ usage reporter used a Google Analytics API that has been deprecated.


alessandrofelder (approved):
I'd suggest renaming the test methods that now don't check reporting-related things as they are now named confusingly?

Maybe something like:
* `testCancelledRunDoesNotReport` ->`testEmptyImageCancels` 
* `testSuccessfulRunReports` -> `testSensibleInputDoesNotCancel`
For both analyse-skeleton and ellipsoid-factor wrapper classes?

But I'll leave that up-to-you to decide before merging @mdoube .

PR #393, opened 2026-01-07
Slice Geometry: add moment weighting mode option

### LINKED ISSUE #392: Pixel-weighted moments in Slice Geometry
Currently, second moments of area, polar moments of inertia and section modulus (_I_, _J_, _Z_) are either strictly geometric (using pixel "area" and distance to axis squared) or have the pixel area weighted linearly relative to user-set min and max values.

To improve biological relevance, it may help to use calibrated density values where these exist instead of pixel 'area'.

JonathanAWilliams:
Related to #392
Adds a “Moment weighting” dropdown and enum parsing in Slice Geometry. Default remains geometric (binary), so there is no behaviour change yet—this is scaffolding for the upcoming pixel-weighted (density/partial area) moment calculations.

Next: implement density-weighted centroid + density-weighted integrand 

image

### DIFF
--- a/Legacy/bonej/src/main/java/org/bonej/plugins/SliceGeometry.java
+++ b/Legacy/bonej/src/main/java/org/bonej/plugins/SliceGeometry.java
@@ -174,6 +174,9 @@ public class SliceGeometry implements PlugIn, DialogListener {
 	private double background;
 	private double foreground;
 	private boolean doPartialVolume;
+	private MomentWeightingMode weightingMode = MomentWeightingMode.GEOMETRIC;
+
+
 
 	@Override
 	public boolean dialogItemChanged(final GenericDialog gd, final AWTEvent e) {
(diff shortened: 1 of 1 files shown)


JonathanAWilliams:
Removed the redundant checkbox as suggested and verified in ImageJ.
Partial-area behaviour is now controlled via the Moment weighting dropdown.
Will start on density-weighted implementaiton 

image

mdoube (dismissed):

JonathanAWilliams:
Added centroid selection driven by the new “Moment weighting” mode.

GEOMETRIC (1): existing behaviour
PARTIAL_AREA (2): reproduces previous partial volume centroid behaviour
DENSITY (3): uses a calibrated per-pixel weight (w = m·pixel + c) to compute a density-weighted centroid

image

In this PR the change is intentionally limited to centroid computation only; the second-moment (integrand) accumulation remains geometric/partial-area as before.

Next step: apply the selected weighting mode inside the second-moment accumulation.

mdoube:
> When enabled, a second dialog allows users to select one or more outputs:
> 
> * Imin contribution
> * Imax contribution
> * J (polar invariant) contribution
> 
> At least one output must be selected to proceed.

Better to have all the options in one dialog. Multiple dependent dialogs create a headache for macro command harvesting and for headless operation. The old dialog style allows you to grey out options if other options are not active, or they can just be there and have no effect if the parent option is not selected.

I don't see your code changes that introduce this option, could you point me to them please?

mdoube:
@JonathanAWilliams I'm doing a bunch of work to make Slice Geometry work as a Modern plugin. Please keep an eye on it and merge the changes back into your weighted moments branch. https://github.com/bonej-org/BoneJ2/pull/400

